PR #67, opened 2021-11-04
Switches back to Steps that are responsible for their own execution and dependencies

dirkgr:
In this PR, I went back to having the `Step` class keep track of its dependencies. Step graphs are represented as `Step` objects that reference each other.
 * It makes the tango-as-a-library ~~easier~~ possible.
 * It means that we don't have to hash all inputs to derive the name of a step. When a step is an input to another step, we only need the step name, not the step's results. This is a requirement for instantiating the step graph without executing it. In practice, it's also a requirement when step results are large and hashing them just to derive a name is not feasible.

I'm a big fan of the new way of parsing step graphs. It have moved around where that happens, but it's clearly better than trying to make `FromParams` do it. It also makes `FromParams ` little simpler, though not as simple as I thought at some point.

I think (but haven't confirmed) that this approach brings back anonymous steps.

We had a problem with complex objects having some of their members initialized from steps. @epwalsh fixed it one way. I fixed it another way in here, by providing the `WithUnresolvedSteps` class, which defers the initialization of the complex object until the step actually runs.

I tried to leave the stuff related to the execution environment (`Executor`, `StepGraph`, `StepCache`, ...) alone, but wasn't able to do so completely. In particular, `Executor` is no longer a registrable, but that concept will have to come back soon. I just didn't want to blow up this PR. I did not keep the concept of the "step group", because I would have had to do work to keep it, and it sounds like Pete wasn't sold on that concept anymore anyways. If we need it I'm happy to bring it back properly.

I renamed the directories a bit, since we used a confusing mess of names. Now there is a "step dir", and a "work dir". The work dir is usually, but not always, inside the step dir. The work dir is scratch space. The step dir is where we store results permanently.


tango/common/from_params.py
```diff
@@ -305,13 +305,31 @@ def construct_arg(
...
     default: Any,
+    could_be_step: bool = True,
     **extras,
 ) -> Any:
     """
     The first two parameters here are only used for logging if we encounter an error.
     """
     from tango.step import Step
 
+    if could_be_step:
```
dirkgr, line 316: This had to come back after all :-/

tango/integrations/torch/train.py
```diff
@@ -295,18 +294,11 @@ def _train(
     is_distributed: bool = False,
     devices: Optional[List[int]] = None,
     distributed_port: str = "54761",
-    include_package: Optional[List[str]] = None,
     val_metric_name: str = "loss",
     minimize_val_metric: bool = True,
     aggregate_val_metric: bool = True,
     callbacks: Optional[List[Lazy[TrainCallback]]] = None,
 ) -> Optional[Model]:
-    if include_package:
```
dirkgr, line 304: Tango already includes packages, no? The `Executor` does that before anything runs? Do we need to do it again?

epwalsh: This is actually important because when we spawn subprocesses (like during distributed training), those processes need to re-import all of the `include_package` modules, or they won't be able to instantiate things.

epwalsh:
> I did not keep the concept of the "step group", because I would have had to do work to keep it, and it sounds like Pete wasn't sold on that concept anymore anyways.

Yeup, "step group" was poorly thought out anyway, so I'm happy to see it go.

epwalsh:
Overall I like these changes. I'll do a full review soon.

epwalsh:
There's still a lot of tests failing, should I hold off on reviewing?

dirkgr:
After a tour de force of making the tests run, this is now ready for review @epwalsh.

dirkgr:
Note: Make sure that changing a default argument changes the unique id of a step.

epwalsh (changes requested):
This is my first pass, I didn't get all the way through a review because I got stuck on a bug which I've added a test case for here: https://github.com/allenai/tango/pull/83

tango/step.py
```diff
@@ -359,15 +352,250 @@ def unique_id(self) -> str:
...
+            if click_logger.isEnabledFor(logging.INFO):
+                message = click.style("\N{check mark} Found output for ", fg="green")
+                message += click.style(f'"{self.name}"', bold=True, fg="green")
+                message += click.style(" in cache", fg="green")
+                if needed_by is None:
+                    message += click.style(" ...", fg="green")
+                else:
+                    message += click.style(f' (needed by "{needed_by.name}") ...', fg="green")
+                click_logger.info(message)
+            return workspace.step_cache[self]
```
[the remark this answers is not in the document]
dirkgr: Then I think we should streamline how the logs are configured. I want it so if you do nothing at all, using Tango as a library, you don't see click logs. But if you use the CLI and you do otherwise nothing, you see click logs by default.

I felt a little uneasy about making these into log messages at all, but I think it has to be that way. We're printing these messages from all parts of the system, so any other solution we come up with, we'd have to plumb it through every function call we have. Or we print it all the time, which breaks the as-a-library case. The early transformers library did this, and they finally, thankfully changed it :-)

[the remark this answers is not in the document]
epwalsh: > Then I think we should streamline how the logs are configured. I want it so if you do nothing at all, using Tango as a library, you don't see click logs. But if you use the CLI and you do otherwise nothing, you see click logs by default.

Agreed. This wasn't an issue before because click messages were only emitted from the Executor, so they could easily be turned off.

tango/__main__.py
```diff
@@ -26,12 +26,14 @@
...
+      --file-friendly-logging         Outputs progress bar status on
+                                      separate lines and slows refresh rate.
       --help                          Show this message and exit.
 
     Commands:
-      info  Get info about the current tango installation.
-      run   Run a tango experiment.
+      info    Get info about the current tango installation
+      run     Run a tango experiment
+      server  Run a local webserver that watches a workspace
```
epwalsh, line 36: Should also add a blurb below about this command, like we have for `tango run` and `tango info`.

tango/step.py
```diff
@@ -49,30 +55,31 @@ def get_args(tp):  # type: ignore
...
       This can be accessed via the :attr:`config` property within each step's :meth:`run()` method.
-    * ``step_executor`` is the :class:`~tango.executor.Executor` being used to run the step.
-      This can be accessed via the :attr:`executor` property within each step's :meth:`run()` method.
     """
 
-    DETERMINISTIC: bool = False
+    DETERMINISTIC: bool = True
     """This describes whether this step can be relied upon to produce the same results every time
     when given the same inputs. If this is ``False``, the step can't be cached, and neither can any
     step that depends on it."""
```
epwalsh, line 85: I don't think this is true. We warn the user but we still allow it.

tango/step.py
```diff
@@ -49,30 +55,31 @@ def get_args(tp):  # type: ignore
...
+
 class Step(Registrable, Generic[T]):
     """
-    This class defines one step in your experiment. To write your own step, just derive from this class
+    This class defines one step in your experiment. To write your own step, derive from this class
     and overwrite the :meth:`run()` method. The :meth:`run()` method must have parameters with type hints.
 
     ``Step.__init__()`` takes all the arguments we want to run the step with. They get passed
     to :meth:`run()` (almost) as they are. If the arguments are other instances of ``Step``, those
     will be replaced with the step's results before calling :meth:`run()`. Further, there are four special
```
epwalsh, line 68: This is not quite true either.

tango/step.py
```diff
@@ -359,15 +352,250 @@ def unique_id(self) -> str:
...
+                ...
+                return MyDataClass(...)
+
+        class ConsumeDataStep(Step):
+            def run(self, input_data: DataWithTimestamp):
+                ...
+
+        produce = ProduceDataStep()
+        consume = ConsumeDataStep(
+            input_data = DataWithTimestamp(produce, time.now()))
```
epwalsh, line 534:
```suggestion
        consume = ConsumeDataStep(
            input_data = DataWithTimestamp(produce, time.now())
        )
```

tango/step.py
```diff
@@ -359,15 +352,250 @@ def unique_id(self) -> str:
...
+
+    That does not work, because :class:`DataWithTimestamp` needs an object of type :class:`MyDataClass`, but we're
+    giving it an object of type :class:`Step[MyDataClass]`. Instead, we change the last line to this:
+
+    .. code-block:: Python
+
+        consume = ConsumeDataStep(
+            input_data = WithUnresolvedSteps(
+                DataWithTimestamp, (produce, time.now()))
```
epwalsh, line 543:
```suggestion
        consume = ConsumeDataStep(
            input_data = WithUnresolvedSteps(
                DataWithTimestamp, produce, time.now()
            )
        )
```

tango/step_graph.py
```diff
@@ -1,107 +1,160 @@
...
-        elif isinstance(o, dict):
-            if set(o.keys()) == {"type", "ref"}:
-                dependencies.add(o["ref"])
+    def __iter__(self) -> Iterator[str]:
+        """
+        The names of the steps in the experiment.
+        """
+        return iter(self.parsed_steps)
+
+    def serialized(self) -> List[Step]:
```
epwalsh, line 143: This name is confusing to me because it could also be interpreted as "serialize" in the context of serialization/deserialization.

tango/local_workspace.py
```diff
@@ -0,0 +1,389 @@
...
+        step_info = self._get_step_info(step)
+        if step_info.status != "running":
+            raise RuntimeError(f"Step {step.name} is failing, but it never started.")
+        step_info.end_time = datetime.now()
+        step_info.error = e
+        self._put_step_info(step, step_info)
+
+    def register_run(self, targets: Iterable[Step], name: Optional[str] = None) -> str:
+        if name is None:
+            name = petname.generate()
```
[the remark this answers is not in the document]
epwalsh: > but I think there might be value in being obviously different from wandb, so people don't confuse the identifiers.

That's fine, but we should make sure our W&B train callback uses *our* generated name as the name of the run in the W&B dashboard. Does the `Step` class have access to that info?

docs/source/api/components/workspace.rst
```diff
@@ -0,0 +1,35 @@
...
+   :members:
+
+Implementations
+---------------
+
+.. autoclass:: tango.workspace.MemoryWorkspace
+   :members:
+
+.. autoclass:: tango.local_workspace.LocalWorkspace
+   :members:
```
epwalsh, line 17: Consider not documenting the members since those are all documented in the base class, and if you don't write new docstrings, the docstrings will be inherited from the base class, which can lead to confusing / misleading docs.
```suggestion
.. autoclass:: tango.workspace.MemoryWorkspace

.. autoclass:: tango.local_workspace.LocalWorkspace
```

docs/source/api/components/step.rst
```diff
@@ -8,6 +8,9 @@ Base class
    :members: 
    :special-members:
```
epwalsh, line 9: Might want to exclude some inherited methods from being documented.
```suggestion
   :special-members:
   :exclude-members: from_params
```

requirements.txt
```diff
@@ -13,6 +13,7 @@ click>=8.0,<9.0
 click-help-colors>=0.9.1,<0.10.0
 tqdm>=4.62,<4.63
 more-itertools>=8.0,<9.0
+petname
```
epwalsh, line 16: Probably a good idea to pin this, even if it's a very loose pin.

epwalsh (changes requested):
This review focuses on documentation.

Also the "Overview" section of the docs will need to be updated. Everything about caching probably needs to be rewritten.

requirements.txt
dirkgr: f13d717

tango/step.py
dirkgr: I don't know why I can't commit these suggestions from the GitHub UI.

docs/source/api/components/step.rst
dirkgr: This is a good idea, but `from_params()` already doesn't show up?

dirkgr: I guess I would like it to generate a concise list of inherited members, with links, rather than duplicating the docs every time.

docs/source/api/components/workspace.rst
dirkgr: That's better, but why do the `StepCache` classes not need that?

tango/step_graph.py
dirkgr: What should we call it then? "`toposort()`"?

tango/local_workspace.py
```diff
@@ -0,0 +1,389 @@
...
+                raise RuntimeError(
+                    f"Step {step.name} is trying to start, but it is already {step_info.status}."
+                )
+
+            step_info.start_time = datetime.now()
+            self._put_step_info(step, step_info)
+        except:  # noqa: E722
+            lock.release()
+            del self.locks[step]
+            raise
```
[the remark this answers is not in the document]
dirkgr: I want the lock to be held if the function succeeds.

tango/local_workspace.py
```diff
@@ -0,0 +1,389 @@
...
+        if step_info.status != "running":
+            raise RuntimeError(f"Step {step.name} is failing, but it never started.")
+        step_info.end_time = datetime.now()
+        step_info.error = e
+        self._put_step_info(step, step_info)
+
+    def register_run(self, targets: Iterable[Step], name: Optional[str] = None) -> str:
+        if name is None:
+            name = petname.generate()
+        run_dir = self.runs_dir / name
```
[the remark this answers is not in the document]
dirkgr: 1d80766003b12391dce1c09d8c92a1b768172196

tango/step.py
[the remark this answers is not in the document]
dirkgr: 19e6de2

tango/local_workspace.py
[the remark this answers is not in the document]
dirkgr: After our discussion, do we still want this? It's not difficult to add a `def run_name() -> Optional[str]` to `Step`, but the name might not be set, and the same step can appear in different runs at different times.

docs/source/api/components/step.rst
epwalsh: I see it here in my local build: http://127.0.0.1:8000/api/components/step.html#tango.step.Step.from_params

tango/step_graph.py
epwalsh: I was thinking `ordered_steps()`

epwalsh:
> There is a general question of what to do with non-cacheable steps in the workspace. We store a whole bunch of metadata for every step execution. Non-cacheable steps can run more than once even in a single run. So I'm inclined to not store anything at all for those steps. It would be a bit of work, and probably not necessary, to store the metadata for all executions. And storing just the latest execution is confusing.

At first thought my preference would be clean the run directory before each run of the step so that it runs as if it's the first time. And of course save all metadata each time.

tango/local_workspace.py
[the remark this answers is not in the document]
epwalsh: Oh, right.

tango/local_workspace.py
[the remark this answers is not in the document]
epwalsh: Since we don't need it now I think we should forget about it. We might have a better solution in the future.

tango/workspace.py
```diff
@@ -0,0 +1,307 @@
...
+    @property
+    def status(self) -> str:
+        """
+        The result will be one of these:
+
+        - ``"incomplete"`` if the step has never started
+        - ``"running"`` if the step is running right now
+        - ``"completed"`` if the step completed successfully
+        - ``"failed"`` if the step failed
+        """
```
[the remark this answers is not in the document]
dirkgr: 7f68a97

tango/step.py
dirkgr: Why not?

tango/step.py
dirkgr: 6db2c29be0699ac3041d750b17356f6c14c003be

tango/__main__.py
dirkgr: 44f79ec2b58efa9bec063c475e761009db3ceccf

tango/step_graph.py
dirkgr: That's not super precise, but good enough. 4e9910d4fe4cdff43e2d1e2a85ebc2c9328fe658

docs/source/api/components/step.rst
dirkgr: That link obviously goes somewhere else on my box. I happen to have the Sphinx server running though, and it goes to a copy that doesn't have `from_params()` 🤪.

I think I'll take the suggestion anyways and we don't think too hard about why it's different on my box until it happens again?

dirkgr:
> At first thought my preference would be clean the run directory before each run of the step so that it runs as if it's the first time. And of course save all metadata each time.

That means the metadata for a non-cached step will always only reflect the last time it ran. Also, this is a problem if we want to run the same non-cached step multiple times in parallel.

In light of the second point, I'm thinking we should make a new step directory every time it runs. Maybe `f"{step.unique_id}-{timestamp}"`? I hope that doesn't blow up any other assumptions anywhere.

My vision for non-cacheable steps was always that they should be so quick and simple that they don't require a bunch of infrastructure to run. They should be little more than a quick function call.

tango/step.py
epwalsh: I think when I first read this I interpreted it as "if the type hint of the argument is `Step` ...", but after reading it again I think it's clear.

epwalsh:
@dirkgr 

> That means the metadata for a non-cached step will always only reflect the last time it ran.

That seems okay to me.

> Also, this is a problem if we want to run the same non-cached step multiple times in parallel.

We acquire locks so this wouldn't be allowed, which also sounds okay to me.

But..

> I'm thinking we should make a new step directory every time it runs

I like that idea as well. I was just worried that it could cause issues internally if the run directory doesn't match the unique ID, but I'm sure there's a way to make it work.

docs/source/api/components/step.rst
epwalsh: Could be different Sphinx versions? I just updated so I'm using the latest. If you've been running the server for a while you may want to shut it down and clean the docs build directory. `make docs` will automatically clean the build directory before starting the server.

dirkgr:
I tried writing this code, but it gets too complicated. Now every call in `Workspace` has two cases, cached step and uncached step. When multiple uncacheable steps run at the same time, the `Workspace` has no way of distinguishing between them. This is getting too complicated for a very moderate gain. Nobody but us will ever be able to write a `Workspace` correctly, and we don't be able to anymore either if we ever step away for six months. I'm going to change it so that the `Workspace` has nothing to do with uncacheable steps at all.

dirkgr:
@epwalsh, this is good for another review.

tango/__main__.py
```diff
@@ -164,22 +160,14 @@ def main(
...
 
-    if no_logging or config.no_logging:
-        config.no_logging = True
-        log_level = None
-        config.log_level = None
-    elif log_level is not None:
-        config.log_level = log_level
-    else:
-        log_level = config.log_level
+    config.log_level = log_level
```
epwalsh, line 167:
```suggestion
    if log_level is not None:
        config.log_level = log_level
```

tango/__main__.py
```diff
@@ -164,22 +160,14 @@ def main(
...
 
     common_logging.initialize_logging(
-        log_level=log_level, file_friendly_logging=file_friendly_logging
+        log_level=log_level, file_friendly_logging=file_friendly_logging, enable_click_logs=True
```
epwalsh, line 170:
```suggestion
        log_level=config.log_level, file_friendly_logging=file_friendly_logging, enable_click_logs=True
```

tango/__main__.py
```diff
@@ -291,53 +304,53 @@ def info(config: TangoGlobalSettings):
...
     step_graph = StepGraph(params.pop("steps", keep_as_dict=True))
-    step_cache = StepCache.from_params(
-        params.pop("cache", default={}), dir=directory / "step_cache"
-    )
-    executor = Executor.from_params(
-        params.pop("executor", default={}),
-        dir=directory,
-        step_cache=step_cache,
+    server = WorkspaceServer.on_free_port(workspace)
+    server.serve_in_background()
```
epwalsh, line 344: I think we should have the option to disable spinning up the server.

epwalsh (changes requested):
This looks really good, I just have a few minor suggestions 👇

tango/__main__.py
dirkgr: That doesn't work though. If `log_level` is `None`, we need it to be `None` in `initialize_logging`, so that it can read out the environment variable. I think we need a different solution altogether.

dirkgr: I put a different solution into 03049fa.

tango/__main__.py
dirkgr: 03049fa

tango/__main__.py
dirkgr: 8d8670a

dirkgr:
All done!

tango/__main__.py
epwalsh: The point of the config was to control the log level (among other things). With https://github.com/allenai/tango/commit/03049fa37c99a5da8b182ba77cb41ff6b6986673 you are just **setting** the config's log level based on the environment variable or the command line argument.

dirkgr:
@epwalsh , is this GTG?

tango/__main__.py
dirkgr: What's the difference between "control the log level" and "set the log level"?

epwalsh (approved):

PR #70, opened 2021-11-05
refactor TorchTrainStep

epwalsh:
I started off thinking I'd work on integrating DeepSpeed, but then realized I needed to refactor the TrainStepFirst.

This should actually change the logic of the train loop in any way, I'm just moving parameter into a new `TrainConfig` dataclass so they are easier to pass around.

### DIFF
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -531,7 +531,6 @@ def _train(
                 callback.pre_batch(step, batch)
             optimizer.zero_grad()
             batch_loss = 0.0
-            batch_outputs: List[Dict[str, Any]] = []
             for micro_batch in batch:
                 # Move tensors to right device.
                 micro_batch = _move_to_device(micro_batch, device)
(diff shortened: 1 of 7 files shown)


tango/integrations/torch/train.py
```diff
@@ -460,25 +412,107 @@ def _train(
...
+            # Link to most recent state path.
+            if config.state_path.is_symlink():
+                config.state_path.unlink()
+            config.state_path.symlink_to(config.state_path_for_step(step))
+
+            # Link to best state path.
+            if is_best_checkpoint():
+                if config.best_state_path.is_symlink():
+                    config.best_state_path.unlink()
+                config.best_state_path.symlink_to(config.state_path_for_step(step))
```
dirkgr, line 483: I thought the best state was saved as a hardlink, not a symlink, so it doesn't break when the file that's named after the number of steps goes away.

dirkgr (approved):
It's a little hard to read because you had to touch so many places, but it seems like a good change. I just hope I didn't miss any changes in functionality.

Can this `TrainConfig` constructed with `FromParams` automatically, through some logic I can't see? If so, we could specify this thing in config files.

tango/integrations/torch/train.py
epwalsh: Good point. Changed.

PR #72, opened 2021-11-10
Disabled Python 3.7

dirkgr:
This turns off Python 3.7 and makes us incompatible with it.

This change will change all step ids, so if you have an existing cache, it will be invalidated.


epwalsh:
We need to use Python 3.7 for the project @AkshitaB and I are working on, because a dependency requires it (https://github.com/bigscience-workshop/promptsource).

dirkgr:
Aw

PR #90, opened 2021-11-24
`SqliteDictFormat`

dirkgr:
Bring back `SqliteDictFormat`, so that we don't have to read huge datasets into memory completely


tango/common/util.py
```diff
@@ -144,4 +152,4 @@ def could_be_class_name(name: str) -> bool:
 
 
 def _is_valid_python_name(name: str) -> bool:
-    return bool(name and name[0].isalpha() and name.isalnum())
+    return bool(name and name[0].isalpha() and name.isalnum())
```
epwalsh, line 155:
```suggestion
    return bool(name and name[0].isalpha() and name.isalnum())

```

tango/format.py
```diff
@@ -346,3 +349,41 @@ def __next__(self) -> T:
...
+        splits = {
+            filename.stem: SqliteSparseSequence(filename, read_only=True)
+            for filename in dir.glob("*.sqlite")
+        }
+        return DatasetDict(metadata=metadata, splits=splits)
+
+    def checksum(self, dir: PathOrStr) -> str:
+        # This is not trivial to implement because sqlite files can be different even if they contain the same
+        # data.
+        raise NotImplementedError()
```
epwalsh, line 389: This makes this format unusable. Maybe we just change the signature of `checksum()` to return `Optional[str]`, and handle that in `StepCache`?

epwalsh: That's fine with me

dirkgr: e54e31f

epwalsh (approved):

PR #101, opened 2021-12-03
Euler

dirkgr:
These are the steps and code from today's demo. Some gnarly fixes in here.


tango/executor.py
```diff
@@ -50,13 +52,35 @@ def execute_step_graph(self, step_graph: StepGraph) -> str:
 
         ordered_steps = sorted(step_graph.values(), key=lambda step: step.name)
 
+        # find uncacheable leaf steps
```
dirkgr, line 55: Turns out that uncacheable leaf steps would never run with the changes I made. That's not right.

dirkgr:
The fact that the style checker in Python 3.7 runs different than in 3.8 or 3.9 really grinds my gears.

dirkgr:
... though actually that wasn't the real problem this time 😶

tango/steps/__init__.py
```diff
@@ -5,5 +5,13 @@
...
 __all__ = ["DatasetRemixStep", "PrintStep"]
 
+from .complex_arithmetic import (
+    AdditionStep,
+    CosineStep,
+    ExponentiateStep,
+    MultiplyStep,
+    SineStep,
+    SubtractionStep,
+)
```
epwalsh, line 15: I don't see these steps being useful outside of examples, so I'd prefer that we don't put them into the `tango` library itself because it adds overhead.

tango/executor.py
epwalsh: I thought that was intentional?

tango/executor.py
```diff
@@ -50,13 +52,35 @@ def execute_step_graph(self, step_graph: StepGraph) -> str:
...
+        interior_steps: Set[Step] = set()
+        for step in ordered_steps:
+            for dependency in step.dependencies:
+                interior_steps.add(dependency)
+        uncacheable_leaf_steps = {
+            step for step in set(step_graph.values()) - interior_steps if not step.cache_results
+        }
+        # also pick any uncacheable direct dependencies of leaf steps
```
epwalsh, line 63: Is that necessary? Won't those dependencies run anyway when we run the steps that depend on them?

epwalsh (changes requested):

tango/executor.py
dirkgr: You are 100% right.

dirkgr:
> I thought that was intentional?

I can see why you would think that. In my mind, we have a large step graph with only a few targets. But of course in practice, with the way that the configuration files work right now, every step is a target, so that distinction becomes confusing.

It would be better if we could separate the concerns of "cacheable" and "side-effect only", but I don't want to burden users with it. I did this because the `"print"` step wouldn't run at the end of an experiment. Maybe it would be better if we instead detect leaf steps that return `None`, and run those regardless of cacheability? I'm not sure that adds enough to warrant the necessary code though.

tango/executor.py
dirkgr: 21f0280

tango/steps/__init__.py
dirkgr: They are the best example of step graphs larger than 3 steps we have right now, so I want to keep them. But maybe they should be part of tests instead.

dirkgr: I moved it to https://github.com/allenai/tango-example.

tango/step_graph.py
```diff
@@ -67,6 +68,11 @@ def __init__(self, params: Dict[str, Params]):
         del done
         del todo
 
+        # make sure default steps are available
+        import tango.steps  # noqa: F401
+
+        import_integrations()
```
epwalsh, line 74: Why did you add this? I'm not a huge fan of unnecessary imports. That's one of my biggest annoyances with AllenNLP.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Sample experiment configurations that prove Euler's identity
```
epwalsh, line 12: I guess this is no longer true.

tango/steps/__init__.py
epwalsh: Personally I think we should keep that example in this repo. It's a good example and you put a lot of work into it. If we keep it here we can add it to the docs and it will be seen by a lot more users. We can also ensure that it's kept up-to-date. We already have the `examples/` folder for things like this.

tango/step_graph.py
dirkgr: Because when you check out https://github.com/allenai/tango-example, and you try to run `tango anything`, it's completely useless. It has the `print` step and nothing else.

CHANGELOG.md
dirkgr: c1c9bd0

tango/step_graph.py
epwalsh: I'm not sure what you mean. I had no issues running that without these lines. Tango already lazily imports steps and integrations automatically when needed.

tango/steps/__init__.py
dirkgr: I brought it back in 23d78247376edf729a73d5eb02afea1537618ffa.

tango/step_graph.py
dirkgr: I have latest tango main, and latest tango-example. In the root directory of -example, I run `tango run euler.jsonnet` without the `--import-package`. I'm only doing this so I get a list of available steps.

It prints
```
tango.common.exceptions.ConfigurationError: 'cmul' not in acceptable choices for type: ['dataset_remix', 'print']
```

I see I have none of the integrations available. Only two steps here.

epwalsh: They are available, that's just a bad error message.

epwalsh: I can fix that in a separate PR

dirkgr: Not that I don't believe you, but how?

epwalsh: https://github.com/allenai/tango/pull/109 fixes.

> Not that I don't believe you, but how?

This function is used to guess which modules need to be imported:
https://github.com/allenai/tango/blob/b2781fd5b410c08546e64919a4228f5ffc365dfa/tango/common/registrable.py#L159

epwalsh: Other than this the PR LGTM

dirkgr: I reverted those changes.

epwalsh (approved):

PR #105, opened 2021-12-06
fix bug with resolving lazy step inputs

epwalsh:
Fixes a bug that happens when a step has a `Lazy` parameter and that `Lazy` object depends on the result from another step.


dirkgr (approved):
Good one!

PR #104, opened 2021-12-06
Add FAQ

schmmd:
Add's an FAQ section with "why tango?".

PR #107, opened 2021-12-06
add FAQs to docs

epwalsh:
Followup to https://github.com/allenai/tango/pull/104.


dirkgr (approved):
It seems nothing is ever as easy as it seems 😞 .

PR #109, opened 2021-12-06
ensure all integrations are imported if we can't find registered name

epwalsh:
See https://github.com/allenai/tango/pull/101#discussion_r763432649

### DIFF
--- a/tango/common/params.py
+++ b/tango/common/params.py
@@ -32,6 +32,7 @@ except ImportError:
 
 from .aliases import PathOrStr
 from .exceptions import ConfigurationError
+from .util import could_be_class_name
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


dirkgr:
This makes it so the steps are there, but it doesn't make them discoverable with `Registrable.list_available`?

epwalsh:
> This makes it so the steps are there, but it doesn't make them discoverable with `Registrable.list_available`?

I'm not sure what you mean. This PR touches the part of the library that handles the automatic yet lazy importing of integrations and other components within the core library. The only thing that will be different after this PR is merged is that when `Registrable` can't find the given`name` in it's registry, it will import _everything_ (all integrations and other components within the core library) so that the error message includes ALL steps (or whatever other registrable class you're dealing with) defined in the core library.

dirkgr:
So this works because `list_available()` gets called after we search. Which solves it in the case of the error message. It just means there is some weirdness when you call `list_available()` yourself somewhere.

🤷🏻‍♂️ good enough. Probably a fair price for not having to import the world every time.

dirkgr (approved):

PR #112, opened 2021-12-14
make DatasetDictBase a MutableMapping

epwalsh:
For consistency with HuggingFace's DatasetDict. This is useful in our project with @ZhaofengWu.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


dirkgr (changes requested):
We might want to implement this in classes that would be hard to make mutable. For example, a dataset that is backed by a GSE bucket. I could be convinced to give a `is_mutable() -> bool` property and throw `NotImplementedError` in the appropriate methods for read-only datasets.

The inner mappings should definitely remain mappings, not dicts. They could be implemented  by something that reads from disk or the network, making it very annoying to implement the whole `Dict` interface.

epwalsh:
I actually don't need this anymore so I'm going to close it for now.

PR #114, opened 2021-12-15
Add DatasetsFormat, making LoadDataset cacheable, fix bug with KeyboardInterrupt

epwalsh:

- Adds `DatasetsFormat` format, which just uses `datasets.save_to_disk` and `datasets.load_from_disk` methods.
- Makes `LoadDataset` step use this new format, but only works for non-streaming datasets.
- Adds a new step `LoadStreamingDataset` for streaming datasets, which is not cacheable.
- Fixes a somewhat major bug caused by `KeyboardInterrupt` exceptions not being caught. This occurred because `KeyboardInterrupt` actually doesn't inherit from `Exception` (https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt). The result is that if a step was interrupted, the workspace would be left in a corrupted state that says the step is still running, even though it has failed.

## Before submitting

## After submitting


dirkgr:
> if a step was interrupted, the workspace would be left in a corrupted state that says the step is still running, even though it has failed.

😳

tango/integrations/datasets/__init__.py
```diff
@@ -63,6 +69,29 @@ def convert_to_tango_dataset_dict(hf_dataset_dict):
...
+    using :meth:`datasets.Dataset.save_to_disk()`.
+
+    It is the default :class:`~tango.format.Format` for the :class:`LoadDataset` step.
+    """
+
+    VERSION = 1
+
+    def write(self, artifact: T, dir: PathOrStr):
+        dataset_path = Path(dir) / "data"
+        artifact.save_to_disk(str(dataset_path))
```
dirkgr, line 88: I'm assuming this writes multiple files, and that's why you stuck it into a directory?

tango/integrations/datasets/__init__.py
```diff
@@ -72,22 +101,65 @@ class LoadDataset(Step):
...
+
+    .. tip::
+
+        Registered as a :class:`~tango.step.Step` under the name "datasets::load_streaming".
+
+    """
+
+    DETERMINISTIC = True
+    VERSION = "001"
+    CACHEABLE = False  # can't be cached
```
dirkgr, line 145: To be sure, Tango can serialize and deserialize iterators. There is support for that in `DillFormat` and I think also `JsonFormat`. But it's not clear that this is necessary in this case.

tango/integrations/datasets/__init__.py
```diff
@@ -72,22 +101,65 @@ class LoadDataset(Step):
...
+    .. important::
+
+        If you are loading an :class:`~datasets.IterableDataset` or :class:`~datasets.IterableDatasetDict`
+        you need to use the :class:`LoadStreamingDataset` step instead.
+
     """
 
     DETERMINISTIC = True
     VERSION = "001"
-    CACHEABLE = False  # These are already cached by huggingface.
```
dirkgr, line 79: To be super clear, we want to do this because even though huggingface caches datasets, their `load_dataset()` function is still sometimes slow enough to make this worth it?

dirkgr: I feel like that may be worth pointing out somewhere in a doc or a comment.

tango/step.py
```diff
@@ -306,7 +306,9 @@ def _run_with_work_dir(self, workspace: "Workspace", **kwargs) -> T:
                 if self.cache_results:
                     result = workspace.step_finished(self, result)
                 return result
-            except Exception as e:
+            except BaseException as e:
+                # TODO (epwalsh): do we want to handle KeyboardInterrupts differently?
```
dirkgr, line 310: I think "failed" is a perfectly acceptable state for this until we find we need to do something special.

tango/integrations/datasets/__init__.py
epwalsh: Yes

tango/integrations/datasets/__init__.py
epwalsh: I think that would be really inefficient, and a streaming dataset could potentially go on forever.

tango/integrations/datasets/__init__.py
epwalsh: Yes, I'll add a comment.

tango/integrations/datasets/__init__.py
epwalsh: 32face0

tango/integrations/datasets/__init__.py
epwalsh: 32face0

tango/integrations/datasets/__init__.py
```diff
@@ -63,6 +69,29 @@ def convert_to_tango_dataset_dict(hf_dataset_dict):
...
+
+    VERSION = 1
+
+    def write(self, artifact: T, dir: PathOrStr):
+        dataset_path = Path(dir) / "data"
+        artifact.save_to_disk(str(dataset_path))
+
+    def read(self, dir: PathOrStr) -> T:
+        dataset_path = Path(dir) / "data"
+        return ds.load_from_disk(str(dataset_path))
```
dirkgr, line 92: Will this load the entire dataset into memory up-front? Or does this open the dataset Arrow-style?

dirkgr (approved):

tango/integrations/datasets/__init__.py
epwalsh: 🏹 style

PR #119, opened 2021-12-16
A step that runs seq2seq LMs in inference mode

dirkgr:
There is so much stuff in here, it's probably easiest to look at the changelog: https://github.com/allenai/tango/pull/119/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed


examples/eval_p3/eval.py
```diff
@@ -0,0 +1,109 @@
...
+
+@dataclass
+class RougeResults:
+    rouge1: RougeScore
+    rouge2: RougeScore
+    rougeL: RougeScore
+
+
+@Step.register("rouge_score")
+class RougeScoreStep(Step[RougeResults]):
```
dirkgr, line 74: We could make this a general step, but it introduces a new dependency.

tango/local_workspace.py
```diff
@@ -374,9 +415,23 @@ def register_run(self, targets: Iterable[Step], name: Optional[str] = None) -> s
         # write step info for all steps
         all_steps = set(targets)
         for step in targets:
-            all_steps.union(step.recursive_dependencies)
-        for step in all_steps:
-            self._put_step_info(step, self._get_step_info(step))
+            all_steps |= step.recursive_dependencies
```
dirkgr, line 418: That was embarrassing.

CHANGELOG.md
```diff
@@ -11,17 +11,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Fixed
 
 - Fixed torch `StepEarlyCallback` state not being recovered properly on restarts.
+- `LocalWorkspace` now works properly with uncacheable steps.
+- When a Tango run got killed hard, with `kill -9`, or because the machine lost power, `LocalWorkspace` would
+  sometimes keep a step marked as "running", preventing further executions. This still happens sometimes, but it
+  is now much less likely (and Tango gives you instructions for how to fix it).
+- To make all this happen, `LocalWorkspace` now saves step info in a Sqlite database. Unfortunately that means that
+  the workspace format changes and existing workspace directories won't work properly with it.
```
epwalsh, line 33: Oof, this is not ideal for Zhaofeng's project if we ever want to upgrade tango. Is it possible to write a script that converts an old workspace into a new one?

epwalsh (changes requested):

examples/eval_p3/requirements.txt
```diff
@@ -0,0 +1 @@
+rouge-score
```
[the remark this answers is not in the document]
dirkgr: Why?

CHANGELOG.md
dirkgr: I will look into making it compatible.

tango/common/logging.py
```diff
@@ -114,3 +116,78 @@ def initialize_logging(
         FILE_FRIENDLY_LOGGING = True
         os.environ["FILE_FRIENDLY_LOGGING"] = "true"
         click_logger.disabled = True
+
+
+def logging_tqdm(
```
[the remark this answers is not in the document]
dirkgr: 1. You don't have to set an environment variable. It's always on.
2. It does not redirect `stderr` in a dodgy way.
3. Log messages get written to your logger, not the global `"tqdm"` logger.
3. It does not depend on implementation details of the original tqdm.
4. It is less code.

examples/eval_p3/config.jsonnet
```diff
@@ -0,0 +1,63 @@
...
+    'xsum_DOC_tldr',
+    'xsum_DOC_write_summary_of_above',
+    'xsum_article_DOC_summary',
+    'xsum_college_roommate_asked_DOC_so_I_recap',
+    'xsum_read_below_DOC_write_abstract',
+    'xsum_summarize_DOC',
+    'xsum_summarize_this_DOC_summary'
+];
+
+local dataset_steps = std.foldl(
```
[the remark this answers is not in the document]
dirkgr: 15e535a

examples/eval_p3/requirements.txt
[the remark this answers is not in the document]
epwalsh: It's helpful for CI so we can install everything with one command. For example, `pip install -e [dev,examples]`. Speaking of, it would be great to have tests for this example just like we have for the `train_gpt2` example. https://github.com/allenai/tango/blob/8e09b66caffbff20fd0b1504c961932b97417e8d/.github/workflows/ci.yml#L81-L85

tango/common/logging.py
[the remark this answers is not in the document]
epwalsh: Let me rephrase. What I'm really asking is this: (A) why not implement this using our existing `Tqdm` wrapper, and (B) why have two separate approaches for file friendly progress bars?

For (A), your 2nd point is still valid, but your approach of implementing a new tqdm from scratch adds more code that looks fairly tricky. We should at least have some test coverage here.

But to emphasize (B) again, it seems like there is a lot of overlap in use-cases here for you `logging_tqdm` and the existing `Tqdm` wrapper with `FILE_FRIENDLY_LOGGING` on. I'd rather go with one or the other. Maybe our file-friendly version of the `Tqdm` wrapper uses your `logging_tqdm` code instead of `tqdm` internals.

examples/eval_p3/eval.py
```diff
@@ -0,0 +1,109 @@
...
+from tango import Format, JsonFormat, Step
+from tango.common import DatasetDict
+from tango.common.logging import make_tqdm
+
+logger = logging.getLogger(__name__)
+tqdm = make_tqdm(logger)
+
+
+@dataclass
+class RougeScore:
```
epwalsh, line 16: This class adds a lot of code to the example that seems overkill for just aggregating a metric. Seems like the `Score` named tuple from `rouge_score.scoring` is good enough already. For example, you could aggregate a bunch of `Score` objects like this:

```python
scores = [Score(0.5, 0.4, 0.6), Score(0.1, 0.3, 0.2)]
Score(*[sum(x) / len(scores) for x in zip(*scores)])
```

tango/common/util.py
```diff
@@ -153,3 +153,34 @@ def could_be_class_name(name: str) -> bool:
...
+        try:
+            for value in g:
+                q.put(value)
+        finally:
+            q.put(sentinel)
+
+    thread = Thread(name=repr(g), target=fill_queue, daemon=True)
+    thread.start()
+
+    yield from iter(q.get, sentinel)
```
epwalsh, line 186: Should we join the thread at the end?

```suggestion
    yield from iter(q.get, sentinel)
    
    thread.join()
```

tango/common/util.py
```diff
@@ -153,3 +153,34 @@ def could_be_class_name(name: str) -> bool:
 
 def _is_valid_python_name(name: str) -> bool:
     return bool(name and name[0].isalpha() and name.isalnum())
+
+
+def threaded_generator(g, queue_size: int = 16):
```
epwalsh, line 158: If you want this in the API docs you can import it in `tango/common/__init__.py`.

CHANGELOG.md
```diff
@@ -11,17 +11,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added `DatasetCombineStep`, a step that combines multiple datasets into one.
 
 ### Fixed
 
 - Fixed torch `StepEarlyCallback` state not being recovered properly on restarts.
+- `LocalWorkspace` now works properly with uncacheable steps.
+- When a Tango run got killed hard, with `kill -9`, or because the machine lost power, `LocalWorkspace` would
+  sometimes keep a step marked as "running", preventing further executions. This still happens sometimes, but it
+  is now much less likely (and Tango gives you instructions for how to fix it).
+- To make all this happen, `LocalWorkspace` now saves step info in a Sqlite database. Unfortunately that means that
```
epwalsh, line 32: > To make all this happen, `LocalWorkspace` now saves step info in a Sqlite database.

I'm curious how using an Sqlite dataset addresses these issues.

CHANGELOG.md
dirkgr: 2cd8994a99c9d2b228ed776186946d6bd3bf29d0

epwalsh: Yay!

CHANGELOG.md
dirkgr: We used to store `StepInfo` in the step directory, but uncacheable steps don't have run directories, so we had no way of storing info for uncacheable steps. The web UI was freaking out about that.

It doesn't have to be sqlite, but that was an easy option. We could also create a directory (outside the cache) with a bunch of JSON files, one for each step. Then at least they would be human-readable, which is a plus.

(Sqlite had nothing to do with automatically detecting hung steps, though in real adversarial `kill -9` scenarios, it's probably helpful that it has transactions.)

CHANGELOG.md
dirkgr: That ended up way bigger than intended because it turns out we had no tests for local workspaces (😳), and to write those tests I had to fix another issue that I wanted to leave for another PR.

tango/common/util.py
dirkgr: Yes. I've been using that function for years and never noticed. Though I checked under the debugger, and by the time we get to that line, the thread is already dead, so maybe that's why.

dcef5b4

tango/common/util.py
dirkgr: dcef5b4

examples/eval_p3/eval.py
dirkgr: I keep three of these score tuples around, and I do four arithmetic operations with them, so I'd end up with 12 lines like `Score(*[sum(x) / len(scores) for x in zip(*scores)])`. Though I can probably condense two of the operations, so then it's 9.

I'd argue that while mine has more lines, it's easier to read. I could remove some of the actions (`__imul__` is never used, for example), but this is not a test, it's an example of how to write code.

examples/eval_p3/requirements.txt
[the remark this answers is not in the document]
dirkgr: I think I will switch this to `torchmetrics`. It's not a new dependency, because it's part of PTL. And as of today, it has Rouge.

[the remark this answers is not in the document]
dirkgr: 87bfc53

But, to make this work, I had to make `"examples"` into an integration. `torchmetrics` is installed with lightning, but I want to guarantee the correct version. Since 0.7.0 came out today, it does not yet get installed by default.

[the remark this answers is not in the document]
epwalsh: Why does `"examples"` have to be an integration? Why not put this new dependency in `dev-requirements.txt` along with other dependencies for examples?
https://github.com/allenai/tango/blob/8e09b66caffbff20fd0b1504c961932b97417e8d/dev-requirements.txt#L54-L57

examples/eval_p3/eval.py
dirkgr: None of this code is still around after switching to torchmetrics.

examples/eval_p3/requirements.txt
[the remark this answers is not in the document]
dirkgr: I put it there.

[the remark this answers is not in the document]
dirkgr: cc6a238

tango/common/logging.py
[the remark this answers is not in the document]
dirkgr: I'm happy to write tests for it, but only if we decide to actually use it.

I looked at our current usage of TQDM, and the only feature that `logging_tqdm` is missing is `wrapattr()`. There are a few others, but those only apply to rendering visual progress bars, so we don't have to care about them.

Overall I think `logging_tqdm()` is the superior solution, but there is a lot to do, and the success of Tango won't hinge on the quality of the progress bars. So I'll take this thing out and replace it with `FILE_FRIENDLY_LOGGING`.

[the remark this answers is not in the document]
dirkgr: c3a743b

dirkgr:
I think the last thing in here is the issue with how to select GPUs. I see how you did it in afade9262f3aa5300824d4dae80ec472e73dccda. There, you made the evaluation step completely automatic, while the training step takes the `device_count` parameter. If we decide to do that here, this PR is done. Sound good @epwalsh?

epwalsh:
> There, you made the evaluation step completely automatic, while the training step takes the `device_count` parameter. If we decide to do that here, this PR is done. Sound good @epwalsh?

Yes, I mean, the only reason the `TorchEvalStep` doesn't take a `device_count` parameter at the moment is because it only works on a single device (for now).

dirkgr:
Same is true for this step. But even if it could run on multiple devices, it would not alter the results. We could still put in a `device_count` parameter, but it would be a `SKIP_ID` parameter.

dirkgr:
Is this good to go then?

dirkgr:
Ah, I have to call `resolve_device()` as soon as #120 is merged.

epwalsh (approved):
Yeup, this looks good other than the merge conflicts and using `resolve_device()` ✅

Issue #116, opened 2021-12-16
`torch::train` step takes `devices` as a parameter


dirkgr:
Since `devices` is an input to `torch::train`, it'll be part of the step's unique id. That means it'll recompute when the devices change. I think it's safe to assume that training a model on 0 GPUs vs. 1 GPU vs. 2 GPUs will produce the same model, so this should not be part of the parameters. We should automatically use all GPUs we can take advantage of and use `CUDA_VISIBLE_DEVICES` to restrict their use.

epwalsh:
> We should automatically use all GPUs we can take advantage of and use CUDA_VISIBLE_DEVICES to restrict their use.

I disagree with that. Moving up to multi-GPU training from single GPU training requires additional care to make sure your data gets properly sharded, and it also changes your effective batch size and the way instances are sampled. So training on multiple GPUs **will** likely produce a different model.

dirkgr:
If I have a total batch of size 32, and I send 16 instances to 2 GPUs each, I should end up with the same model afterwards, no? It should work just like gradient accumulation, except fast.

epwalsh:
> If I have a total batch of size 32, and I send 16 instances to 2 GPUs each, I should end up with the same model afterwards, no?

Ideally, yes, but in practice probably not. It might be tricky to set your RNG states correctly to guarantee you get the same results, and I think you'd also have differences with `LayerNorm`.

But my point about batch size is that we treat the `batch_size` parameter as the batch size per GPU.

dirkgr:
Layer norm would be fine, but batch norm would not.

Fair enough. So if we take that to the logical conclusion, we have the opposite problem: `devices=None` will produce a different outcome with the same parameters, and thus with the same unique ID, depending on machine you run it on. I think we should switch it up so `devices` is required, but in the step's `__init__()` method we autodetect devices and fill in the parameter if we have to.

epwalsh:
Actually I think the current behavior when `devices=None` is okay. When `devices=None` we train on a single GPU if one is available, otherwise we train on CPU. We never train on multiple GPUs unless you explicitly set `devices`. And it's probably reasonable to assume that you get the same model if you train on a single GPU or CPU only.

PR #120, opened 2021-12-16
Add TorchEvalStep to torch integration, use "device_count" in TorchTrainStep instead of "devices"

epwalsh:

- Adds a step for evaluating a model to the torch integration: `TorchEvalStep`, registered as "torch::eval".
- Renames `aggregate_val_metric` to `auto_aggregate_val_metric` in `TorchTrainStep`.
- Replaces `devices` parameter of `TorchTrainStep` with `device_count`. #116 

## Before submitting

## After submitting


tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+        If this is not correct for your metric you will need to handle the aggregation
+        internally in your model or with an :class:`EvalCallback`
+        using the :meth:`EvalCallback.post_batch()` method.
+        Then set the parameter ``auto_aggregate_metrics`` to ``False``.
+
+    """
+
+    DETERMINISTIC = True
+    CACHEABLE = True
+    FORMAT: Format = JsonFormat()
```
epwalsh, line 47: MyPy always complains about this one if you don't annotate `FORMAT`. Idk why.

dirkgr: I have noticed this, but I have long ago given up trying to understand mypy...

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+    def run(  # type: ignore[override]
+        self,
+        model: Model,
+        dataset_dict: DatasetDictBase,
+        dataloader: Lazy[DataLoader],
+        test_split: str = "test",
+        seed: int = 42,
+        eval_steps: Optional[int] = None,
+        log_every: int = 1,
+        device: Optional[Union[int, str, torch.device]] = None,
```
dirkgr, line 58: Do we expect different results when the eval is run on a different device? I know we had this discussion for training. Does that also apply to evaluations?

Do we expect different results if the step runs on device 1 vs. device 2? Because that's how Tango will treat this.

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+                _device = torch.device(f"cuda:{device}")
+            else:
+                _device = torch.device("cpu")
+        elif isinstance(device, str):
+            _device = torch.device(device)
+        elif isinstance(device, torch.device):
+            _device = device
+        else:
+            raise ValueError(f"unexpected type for 'device': '{device}'")
+        device: torch.device = _device
```
dirkgr, line 130: We should settle on a way to handle devices, and then put this into a utility function as well.

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+        dataloader: DataLoader = dataloader.construct(dataset=dataset_dict[test_split])
+
+        steps: int
+        try:
+            dataloader_len = len(dataloader)
+            steps = dataloader_len if eval_steps is None else min(dataloader_len, eval_steps)
+        except TypeError:
+            if eval_steps is None:
+                raise ConfigurationError(
+                    "You must set 'eval_steps' for streaming/iterable datasets"
```
dirkgr, line 145: I really wanted to get rid of streaming datasets. They make everything more complicated, duplicate every code path, and are usually not necessary. Lazy loading solves the problem 90% of the time.

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+        elif isinstance(device, str):
+            _device = torch.device(device)
+        elif isinstance(device, torch.device):
+            _device = device
+        else:
+            raise ValueError(f"unexpected type for 'device': '{device}'")
+        device: torch.device = _device
+
+        # Prep model.
+        model = model.eval().to(device)
```
dirkgr, line 133: Iz believes with some huggingface models you have to set `required_grad` to `False` manually as well: https://github.com/allenai/tango/pull/119/files#diff-e729d4871ba6a31ef859c8dd9a7400a00e69d736ef9e7e8630bc795378c1a0f5R134

I had some code that OOMed without this, but then worked when I put this in. But I also put in other stuff at the same time, so I don't have 100% confirmation that this is necessary.

epwalsh: I don't think that's necessary anymore when you use `with torch.inference_mode()` instead of `with torch.no_grad()`.

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+        self,
+        model: Model,
+        dataset_dict: DatasetDictBase,
+        dataloader: Lazy[DataLoader],
+        test_split: str = "test",
+        seed: int = 42,
+        eval_steps: Optional[int] = None,
+        log_every: int = 1,
+        device: Optional[Union[int, str, torch.device]] = None,
+        metric_names: Sequence[str] = ("loss",),
```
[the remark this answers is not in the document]
epwalsh: Yeup

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,209 @@
...
+        callbacks : ``List[TrainCallback]``
+            A list of :class:`EvalCallback`.
+
+        """
+        # Set seeds.
+        random.seed(seed)
+        np.random.seed(seed)
+        torch.manual_seed(seed)
+        if torch.cuda.is_available():
+            torch.cuda.manual_seed_all(seed)
```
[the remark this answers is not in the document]
epwalsh: Done. f4bd9f4

tango/integrations/torch/eval.py
epwalsh: Ideally it shouldn't matter, but it might (CPU vs GPU). But if you leave `device` as `None`, tango will automatically decide between GPU and CPU, using GPU whenever there is one available. I don't know if there's a reliable way to choose *which* GPU to use when there is more than one available.

tango/integrations/torch/eval.py
epwalsh: 7efcede

tango/integrations/torch/eval.py
epwalsh: I disagree. There are still scenarios where streaming datasets are useful. For one, your data might actually come from a stream. Another case I ran into recently is when I wanted to combine a bunch of different datasets into a multi-task mixture, but some datasets needed to be undersampled because they were too large. It was easier to implement this as a streaming dataset.

dirkgr: What example do we have of datasets that are truly only available in stream form?

Mixing datasets should be easy with lazily loaded datasets. Why was it not?

tango/integrations/torch/eval.py
dirkgr: :+1:

tango/integrations/torch/eval.py
dirkgr: If we assume that CPU and GPU training are different, we can't also auto-detect whether we want GPU support or not. Or at least, the auto-detection has to happen before the step gets instantiated. Otherwise, when you run with `device=None`, you'll get different results depending on which machine you run on, but the same unique id.

If we assume that CPU and GPU training are the same, we should not have a `device` parameter at all.

I think we can assume that running on one GPU is the same as running on another. I know this is not strictly true, but it's true enough.

I would argue that the difference between CPU and GPU also falls into this category: Not strictly the same, but same enough for practical purposes. But I'm willing to be convinced.

epwalsh: The way it is right now leaves this decision up to the user. If they leave `device=None`, they are telling tango that they will get the same result on GPU or CPU. Otherwise they can explicitly `device` to "0" or "cpu", whatever they want.

tango/integrations/torch/eval.py
epwalsh: > Mixing datasets should be easy with lazily loaded datasets. Why was it not?

Take a look at https://github.com/allenai/meta-learn-prompt/blob/main/meta_learn_prompt/data/mixer_dataset.py to see how complex it got.

> What example do we have of datasets that are truly only available in stream form?

I don't have any actual examples, but I'm sure they're out there.

We already have support for streaming datasets, it really doesn't add much complexity in my opinion. So I don't understand why we'd want to get rid of it.

dirkgr:
Didn't we decide to have people give one parameter, `device_count` or something like that, instead of having them specify exactly which devices they want to use?

epwalsh:
Yes, I haven't got to that yet. I'll ping you when I do.

epwalsh:
Didn't we decide to have people give one parameter, device_count or something like that, instead of having them specify exactly which devices they want to use?

@dirkgr afade92

dirkgr:
Is this ready for a review?

epwalsh:
Yeup

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,187 @@
...
+        ]
+        for callback in callbacks:
+            callback.pre_eval_loop()
+
+        eval_batches = enumerate(islice(dataloader, steps))
+
+        running_metrics: Dict[str, float] = defaultdict(float)
+        aggregated_metrics: Dict[str, float] = {}
+
+        with Tqdm.tqdm(eval_batches, desc="Evaluating", total=steps) as batch_iter:
```
dirkgr, line 147: You could use `threaded_generator` here. It might make things faster, depending on how long the data loader takes. The problem is, when something goes wrong, it's hard to debug the stack traces you get. Maybe for these kinds of steps we should err on the side of debuggability.

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,187 @@
...
+                    for k in metric_names
+                }
+
+                # Aggregate metrics.
+                if auto_aggregate_metrics:
+                    for k in batch_metrics:
+                        running_metrics[k] += batch_metrics[k]
+                        aggregated_metrics[k] = running_metrics[k] / (step + 1)
+                else:
+                    aggregated_metrics.update(batch_metrics)
```
dirkgr, line 173: I'm pretty excited about torchmetrics' latest release. It is not the right time to integrate that here now, but torchmetrics would make this go away (because all metrics already know how to aggregate). This works even across GPUs, which does not work here.

tango/integrations/torch/train.py
```diff
@@ -180,24 +189,19 @@ def run(  # type: ignore[override]
...
-            elif all((x >= 0 for x in devices)):
-                if torch.cuda.device_count() < len(set(devices)):
-                    raise ConfigurationError(
-                        f"Only found {torch.cuda.device_count()} CUDA devices, "
-                        f"but you specified {len(set(devices))} device IDs"
-                    )
-            elif not all((x == -1 for x in devices)):
-                raise ConfigurationError("Invalid value for 'devices'")
+        if device_count <= 0:
+            raise ConfigurationError("Invalid value for 'device_count'. Must be at least 1.")
```
dirkgr, line 193:
```suggestion
            raise ConfigurationError("Invalid value for 'device_count'. Must be at least 0.")
```

tango/integrations/torch/train_callback.py
```diff
@@ -55,6 +55,13 @@ def __init__(
         self.validation_dataloader = validation_dataloader
         self.lr_scheduler = lr_scheduler
 
+    @property
+    def step_id(self) -> str:
+        """
+        The unique ID of the current :class:`~tango.Step`.
+        """
+        return self.train_config.step_id
```
dirkgr, line 63: I don't have the overview here, but will this work if we don't use configuration files, and thus we don't have a config?

tango/integrations/torch/util.py
```diff
@@ -0,0 +1,79 @@
...
+    torch.manual_seed(seed)
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed_all(seed)
+
+
+def resolve_device(device: Optional[Union[int, str, torch.device]] = None) -> torch.device:
+    if device is None:
+        if torch.cuda.is_available():
+            # TODO (epwalsh, dirkgr): automatically pick which GPU to use when there are multiple
+            print("CUDA is available")
```
dirkgr, line 65: We probably don't want to `print()`?

dirkgr (changes requested):
No major changes requested. Just some leftovers.

tango/integrations/torch/train.py
epwalsh: Mm no I think this is right how it is. You can't have 0 devices. 1 device means use a single GPU if available, otherwise use CPU. Otherwise, if we treated 0 devices as "use CPU", then the unique ID of the step would be different for CPU vs GPU, which we decided was not the right behavior.

tango/integrations/torch/train_callback.py
epwalsh: Yea, the `train_config` object is not a  configuration file, it's just an internal object we pass around here.

tango/integrations/torch/util.py
epwalsh: Yeup, good catch. 87d738a

tango/integrations/torch/eval.py
```diff
@@ -0,0 +1,187 @@
...
+        dataset_dict : :class:`~tango.common.dataset_dict.DatasetDictBase`
+            Should contain the test data.
+        dataloader : :class:`DataLoader`
+            The data loader that generates test batches. The batches should be :class:`dict`
+            objects.
+        test_split : :class:`str`, optional
+            The name of the data split used for evaluation in the ``dataset_dict``.
+            Default is "test".
+        seed : :class:`int`, optional
+            Used to set the RNG states at the beginning of training.
```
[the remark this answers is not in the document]
epwalsh: Good catch 87d738a

tango/integrations/torch/eval.py
epwalsh: I know there are plenty of downsides to streaming datasets, but my whole argument here is that someone, somewhere may want to use streaming datasets, so we might as well support them where we can especially since PyTorch and HF datasets officially support them too.

epwalsh: I'm not saying we bend over backwards to support streaming datasets **everywhere**, just that we might as well support them where it's easy to. And I think it's easy to support them in `TorchEvalStep` and in `TorchTrainStep`.

tango/integrations/torch/train.py
dirkgr: Yes, you are right.

tango/integrations/torch/eval.py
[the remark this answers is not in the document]
dirkgr: If the step does nothing random, we don't need to set the seed at all.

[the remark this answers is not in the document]
epwalsh: Oh, I read your first comment wrong. Well the dataloader might have some randomness in it (padding noise, shuffling, etc).

tango/integrations/torch/eval.py
dirkgr: This is not the PR to address it, but streaming datasets are among the worst design choices that the PyTorch team made, and there is considerable competition for that prize. It would be much lower mental overhead if users could know that in Tango, all datasets are `DatasetDict`, and all of them support random access. It might take some code behind the scenes to realize the dream, but we are in a great place to be providing that kind of code.

dirkgr (approved):
I think this is done then?

PR #132, opened 2022-01-07
Store log output to a file in run directory, other logging improvements

### LINKED ISSUE #65: Executor should save logs for each step to a file in the step run dir

epwalsh:
Closes #65 and additional logging improvements.

- `tango run ...` results an `out.log` file being written the run directory. This will include all log records, exceptions, warnings, and tqdm output (from our `Tqdm` wrapper, but we can't control general tqdm output unless we patch stderr/stdout, but I'd rather not go there).
- Added `initialize_worker_logging()` function for properly configuring logging in worker processes/threads.  This uses a socket to send log records back to the main process to avoid race conditions on the `out.log` file. In the future we could also use this socket to send log records over the network, which could be useful for remote workspaces / remote step execution.
- Ensured exceptions and warnings (from the `warnings` module) will be captured in the logs.
- Improved file friendly logging by stripping special styling characters.

## Tips for reviewers

The meat of the changes and new features is all in `tango/common/logging.py`, so that's where you should focus your review. Here's a quick command you can run to see the new functionality in action:

```
tango --log-level info run test_fixtures/experiment/multiprocessing.jsonnet -i test_fixtures.package -d /tmp/tango
```

Then check out the log file created:

```
cat /tmp/tango/runs/*/out.log
```

## Before submitting

## After submitting


tango/__main__.py
```diff
@@ -105,6 +108,14 @@ class TangoGlobalSettings(FromParams):
     By default, it is set to ``False``.
     """
 
+    multiprocessing_start_method: str = "spawn"
```
epwalsh, line 111: Turns out the start method needs to be set as early as possible, and only "spawn" and "forkserver" work with PyTorch DDP on GPU.

tango/common/logging.py
```diff
@@ -1,72 +1,196 @@
...
 
-    def info_once(self, msg, *args, **kwargs):
-        if msg not in self._seen_msgs:
-            self.info(msg, *args, **kwargs)
-            self._seen_msgs.add(msg)
 
-    def warning_once(self, msg, *args, **kwargs):
-        if msg not in self._seen_msgs:
-            self.warning(msg, *args, **kwargs)
-            self._seen_msgs.add(msg)
```
epwalsh, line 33: We've never used these, so I removed them.

tango/common/logging.py
```diff
@@ -1,72 +1,196 @@
...
+
+_LOGGING_QUEUE: Optional[mp.Queue] = None
+"""
+Used to send log records from worker processes back to the main logging thread.
+"""
+
+_LOGGING_THREAD: Optional[threading.Thread] = None
+"""
+Thread used for logging records from worker processes.
+"""
```
epwalsh, line 124: This queue and thread business was inspired by https://docs.python.org/dev/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes

epwalsh: It's necessary to safely write logs from worker processes to the same file.

tango/common/logging.py
```diff
@@ -113,4 +305,140 @@ def initialize_logging(
...
+
+    # Write uncaught exceptions to the logs.
+    def excepthook(exctype, value, traceback):
+        # For interruptions, call the original exception handler.
+        if issubclass(exctype, (KeyboardInterrupt, SigTermReceived)):
+            sys.__excepthook__(exctype, value, traceback)
+            return
+        root_logger.critical("Uncaught exception", exc_info=(exctype, value, traceback))
+
+    sys.excepthook = excepthook
```
epwalsh, line 336: Borrowed from allennlp

epwalsh:
I'm downgrading this to a draft because I have an idea to improve this further.

tango/common/util.py
```diff
@@ -153,3 +159,16 @@ def could_be_class_name(name: str) -> bool:
 
 def _is_valid_python_name(name: str) -> bool:
     return bool(name and name[0].isalpha() and name.isalnum())
+
+
+def find_open_port() -> int:
```
epwalsh, line 164: Borrowed from AllenNLP, so this is actually pretty well tested.

tango/integrations/wandb/torch_train_callback.py
```diff
@@ -65,9 +65,9 @@ def __init__(
         super().__init__(*args, **kwargs)
 
         if self.is_local_main_process and "WANDB_API_KEY" not in os.environ:
-            print(
+            warnings.warn(
```
epwalsh, line 68: Better to use the `warnings` module since these will be redirected to the logging system anyway when logging is configured.

tango/common/util.py
dirkgr: The problem is that this is not thread safe. Two tango processes starting up at the same time might still attempt to start on the same port, because there is a gap between `find_open_port()` and binding to the port. This could be more likely than you think if someone starts tango processes from a script, which would have them go through startup at exactly the same time.

Can you just try ports when the server starts up, the way the web server does?

tango/executor.py
```diff
@@ -41,9 +41,19 @@ def execute_step_graph(self, step_graph: StepGraph) -> str:
...
+            run_name = self.workspace.register_run(
+                step for step in step_graph.values() if step.cache_results
+            )
+        else:
+            try:
+                self.workspace.registered_run(run_name)
+            except KeyError:
+                self.workspace.register_run(
+                    (step for step in step_graph.values() if step.cache_results), name=run_name
+                )
```
dirkgr, line 54: If the run with that name already exists, this will do nothing? That won't work. The workspace has to know about the target steps that are in the run. If we have another run with the same steps, it'll need to be updated.

There is also a risk that you start two runs with the same name but different steps, which would write over each other. That's why runs are write-once so far. I could see an argument for making runs in a workspace more mutable, but then we have to be ready for concurrent modification of runs.

I think it would be easier if we number runs with duplicate names. I.e., you end up with `flying-weasel`, `flying-weasel-2`, `flying-weasel-3`, etc.

tango/step.py
```diff
@@ -176,6 +175,13 @@ def __init__(
 
         self._config = step_config
 
+    @property
+    def logger(self) -> logging.Logger:
+        """
+        A :class:`logging.Logger` that can be used within the :meth:`run()` method.
+        """
+        return logging.getLogger(self.__class__.__name__)
```
dirkgr, line 183: Why is it better to have this as a property instead of as a member?

dirkgr (changes requested):
Unless there is some transparent TCP connection re-use, I suspect this is really expensive. Most messages will be on the order of 100 bytes, and every time it creates and tears down a TCP connection.

That said, I don't think you should work on performance at this point. Let's see if it ever becomes a problem.

tango/__main__.py
```diff
@@ -357,25 +395,31 @@ def _run(
...
+    run_name = workspace.register_run(step for step in step_graph.values() if step.cache_results)
+    run_dir = workspace.run_dir(run_name)
+
+    # Capture logs to file.
+    with common_logging.file_handler(run_dir / "out.log"):
+        # Initialize server.
+        server = None
+        if start_server:
+            server = WorkspaceServer.on_free_port(workspace)
+            server.serve_in_background()
```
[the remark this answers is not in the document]
epwalsh: That wouldn't be too hard, we'd just have to set `propagate` to `False` for that logger so its log records aren't sent up to the root logger.

Does the server produce any logs at the moment? I don't see any at INFO level.

tango/common/logging.py
```diff
@@ -1,116 +1,527 @@
...
+            slen = struct.unpack(">L", chunk)[0]
+            chunk = self.connection.recv(slen)
+            while len(chunk) < slen:
+                chunk = chunk + self.connection.recv(slen - len(chunk))
+            obj = self.unPickle(chunk)
+            record = logging.makeLogRecord(obj)
+            self.handleLogRecord(record)
+
+    def unPickle(self, data):
+        return pickle.loads(data)
```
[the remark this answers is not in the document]
epwalsh: I don't think so, not without implementing our own version of `logging.handlers.SocketHandler`.

tango/step.py
epwalsh: When you set it as a member, the logger will get initialized before `initialize_logging()` is called. I was worried this would cause issues. But maybe it's okay.

tango/common/logging.py
```diff
@@ -1,116 +1,527 @@
...
+    enable_click_logs: Optional[bool] = None,
     file_friendly_logging: Optional[bool] = None,
-    prefix: Optional[str] = None,
+    worker_rank: Optional[int] = None,
+    main_process: bool = True,
 ):
-    global FILE_FRIENDLY_LOGGING
-    global TANGO_LOG_LEVEL
+    global FILE_FRIENDLY_LOGGING, TANGO_LOG_LEVEL, TANGO_CLICK_LOGGER_ENABLED
+    global _LOGGING_HOST, _LOGGING_PORT, _LOGGING_SERVER, _LOGGING_SERVER_THREAD
```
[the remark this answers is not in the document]
epwalsh: I think it's convenient this way when other modules need to reference these values. But I guess that doesn't apply to internal vars here like `_LOGGING_HOST`.

tango/common/logging.py
```diff
@@ -1,116 +1,527 @@
...
-    def error_once(self, msg, *args, **kwargs):
-        if msg not in self._seen_msgs:
-            self.error(msg, *args, **kwargs)
-            self._seen_msgs.add(msg)
 
-    def critical_once(self, msg, *args, **kwargs):
-        if msg not in self._seen_msgs:
-            self.critical(msg, *args, **kwargs)
-            self._seen_msgs.add(msg)
+class WarningFilter(logging.Filter):
```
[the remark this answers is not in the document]
epwalsh: Not that I can find

tango/__main__.py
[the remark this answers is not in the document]
dirkgr: I think the Python web server components have their own logging. It's quite noisy when you run with log level INFO.

tango/executor.py
epwalsh: Hmm yeup, good point. I did this because I need to know the run directory before executing the step graph in order to initialize logging. I'll have to rethink this.

epwalsh: Fixed in e21af7f

epwalsh:
> The problem is that this is not thread safe. Two tango processes starting up at the same time might still attempt to start on the same port, because there is a gap between find_open_port() and binding to the port. This could be more likely than you think if someone starts tango processes from a script, which would have them go through startup at exactly the same time.
>
> Can you just try ports when the server starts up, the way the web server does?

Done: b69e8f1

epwalsh:
> They seem pretty useful though, and they're part of the API we offer.

Okay, fine: b9fd1bf

epwalsh:
> Unless there is some transparent TCP connection re-use, I suspect this is really expensive. Most messages will be on the order of 100 bytes, and every time it creates and tears down a TCP connection.

I'm pretty sure there is connection re-use. See https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SocketHandler. In particular, from the [`emit()`](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SocketHandler.emit) method: "If the connection was previously lost, re-establishes the connection.".

tango/common/logging.py
```diff
@@ -43,74 +178,387 @@ def critical_once(self, msg, *args, **kwargs):
...
 
 
 click_logger = logging.getLogger("click")
+"""
+A logger that logs messages through
+`click <https://click.palletsprojects.com/>`_'s
+``click.echo()`` function.
+
+This is provides a convenient way for command-line apps to log pretty, styled messages.
+"""
```
dirkgr, line 292: It should have been my responsibility to write this.

tango/common/logging.py
```diff
@@ -43,74 +178,387 @@ def critical_once(self, msg, *args, **kwargs):
...
+
+    _initialize_logging(
+        log_level=log_level,
+        enable_click_logs=enable_click_logs,
+        file_friendly_logging=file_friendly_logging,
+        main_process=is_main_process,
+    )
+
+
+def initialize_worker_logging(worker_rank: Optional[int] = None):
```
dirkgr, line 360: Since we can autodetect whether we're a worker or a main process, can't we just go down the correct code path automatically, so users only have to call `initialize_logging()` regardless of whether they are workers or main processes?

tango/common/util.py
dirkgr, line 164: Now that you changed how server startup works, do we still need this?

dirkgr (approved):
I have only two comments. Address them or not, it's good to go either way.

This is a really great addition to Tango. Logging is never sexy, but so necessary. When we add network support on top of it (I think all that's missing is communicating IP addresses to workers?), the PyTorch team will steal it and make it part of `torchrun`.

tango/common/logging.py
dirkgr: :+1:

tango/common/util.py
epwalsh: Yes, this is called in a loop until an open port is successfully claimed. https://github.com/allenai/tango/blob/15ea0ad4eda27c29df41b864e4d5b08a436c3313/tango/common/logging.py#L445-L452

dirkgr: I clearly did not read this carefully enough before. I thought you just did what I did and counted up ports. But you just pass `0` for the port and let the OS handle it! Brilliant!

I just tried this with `socketserver.TCPServer`, and found that you can pass a `0` there too and it'll pick a port. `TCPServer` is the base class of `LogRecordSocketReceiver`, so I think it'll work there too. No utilities required at all.

epwalsh: Oh, cool. I'll try that.

epwalsh: 96a283a

PR #140, opened 2022-01-14
Run without an executor

dirkgr:


dirkgr:
This is now folded into https://github.com/allenai/tango/pull/119.

PR #147, opened 2022-01-20
rename settings.json -> .workspace-metadata.json

epwalsh:


dirkgr (commented):
Do we really need all that stuff? `from_params()`, `to_params()`, data classes, `ClassVar` ... all we're doing is storing an integer. What does it buy us?

epwalsh:
> Do we really need all that stuff? from_params(), to_params(), data classes, ClassVar ... all we're doing is storing an integer. What does it buy us?

You said yourself "I thought maybe we'd put other settings in there at some point." This makes it easier to add more fields in the future.

dirkgr:
Let's make it complicated when we have enough things to store to make it worth it.

epwalsh:
The other reason I did it that way was to be consistent with all of the other `*Metadata` classes. Seems odd to make a special case here for `WorkspaceMetadata` just because we don't have more fields *yet*.

epwalsh:
It's not exactly what `FromParams` was designed for, but IMO the other best alternative for these `*Metadata` classes involves adding another dependency like `pydantic` or `attrs`.

PR #150, opened 2022-01-20
Finetune resnet Example for Tango

SabhyaC26:
Fixes #

- A new example in which we finetune a resnet50 model

## Before submitting

## After submitting


dirkgr:
I think it autocorrected "resnet" to "resent" :-P

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,81 @@
...
+class ResNetWrapper(Model):
+    def __init__(self, num_classes: int, feature_extract: bool, use_pretrained: bool):
+        super().__init__()
+        self.model_ft = models.resnet18(pretrained=use_pretrained)
+        set_parameter_requires_grad(self.model_ft, feature_extract)
+        num_ftrs = self.model_ft.fc.in_features
+        self.model_ft.fc = nn.Linear(num_ftrs, num_classes)
+        self.loss_fn = nn.CrossEntropyLoss()
+
+    def forward(self, image, label) -> Dict[str, torch.Tensor]:
```
epwalsh, line 33: Let's add type hints and make `label` optional so that we can use this model in inference.

```suggestion
    def forward(self, image:  torch.Tensor, label: Optional[torch.Tensor] = None) -> Dict[str, torch.Tensor]:
```

You have to handle creating predictions with `argmax` below when `label` is `None`.

examples/finetune_resnet/config.jsonnet
```diff
@@ -0,0 +1,67 @@
+local data_dir = "./data/hymenoptera_data/";
```
epwalsh, line 1: These examples should be self-contained. Is there a dataset on HuggingFace that we could use instead? Maybe one of these: https://huggingface.co/datasets?task_ids=task_ids:image-classification&sort=downloads

epwalsh (changes requested):
Looks great so far! Just a few minor requests.

Also, I think it would be great to show how to take the trained model and use it for predictions. Maybe you could add another step to the config that takes the trained model and a single image as input and generates the prediction?

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,99 @@
...
+    def forward(self, image: torch.Tensor, label: torch.Tensor) -> Dict[str, torch.Tensor]:
+        output = self.model_ft(image)  # tensor
+        loss = self.loss_fn(output, label)
+        return {"loss": loss}
+
+# Custom data collator for images, that takes in a batch of images and labels and
+# reformats the data so that it is suitable for the model.
+@DataCollator.register("image_collator")
+class ImageCollator(DataCollator[Tuple[torch.Tensor]]):
+    def __call__(self, batch: List[Tuple[torch.Tensor]]) -> Dict[str, Any]:
```
epwalsh, line 43: I think this should be:
```suggestion
    def __call__(self, batch: List[Tuple[torch.Tensor, int]]) -> Dict[str, Any]:
```

SabhyaC26: yup makes sense, changing it

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,122 @@
...
+        def image_loader_wrapper(example_batch):
+            return image_loader(example_batch, input_size=input_size, transform_type="train")
+
+        dataset = dataset.with_transform(image_loader_wrapper)
+        train_test = dataset["train"].train_test_split(test_size=test_size)
+        return train_test
+
+
+@Step.register("prediction")
+class Prediction(Step):
```
epwalsh, line 111:
```suggestion
class Prediction(Step):
    FORMAT: Format = JsonFormat()
```

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,122 @@
...
+        if feature_extracting:
+            for param in model.parameters():
+                param.requires_grad = False
+
+    def forward(
+        self, image: torch.Tensor, label: Optional[torch.Tensor]
+    ) -> Dict[str, torch.Tensor]:
+        output = self.model_ft(image)
+        if label is None:
+            return output
```
epwalsh, line 39: Should convert to the label and return the label instead

SabhyaC26: done

examples/finetune_resnet/config.jsonnet
```diff
@@ -0,0 +1,85 @@
...
+local pretrained_model = "resnet_ft";
+<<<<<<< HEAD
+local training_steps = 500;
+local validate_every = 50;
+local image_url = "https://tinyurl.com/2p9xjvn9";
+=======
+local training_steps = 100;
+local validate_every = 20;
+local image_url = "https://www.google.com/url?sa=i&url=https%3A%2F%2Fwww.countryliving.com%2Flife%2Fkids-pets%2Fa28423224%2Fdog-instagram-captions%2F&psig=AOvVaw1GHVmz_Ftm9Q0wiOWv8RUh&ust=1643836699366000&source=images&cd=vfe&ved=0CAsQjRxqFwoTCKCarIa33_UCFQAAAAAdAAAAABAF";
+>>>>>>> b697284d64614c466c2e1ee273c3e095044db522
```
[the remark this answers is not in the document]
SabhyaC26: fixed on mine

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,138 @@
...
+        return "cat"
+    else:
+        return "dog"
+
+
+@Step.register("prediction")
+class Prediction(Step):
+    FORMAT: Format = JsonFormat()
+
+    def run(self, image_url: str, input_size: int, model: models) -> Dict[str, Any]:  # type: ignore
```
epwalsh, line 127: This should also take a device parameter.
```suggestion
    def run(self, image_url: str, input_size: int, model: models, device: Optional[str] = "cpu") -> Dict[str, Any]:  # type: ignore
        device = torch.device(device)
        model = model.to(device)
```

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,138 @@
...
+            for param in model.parameters():
+                param.requires_grad = False
+
+    def forward(
+        self, image: torch.Tensor, label: Optional[torch.Tensor] = None
+    ) -> Dict[str, torch.Tensor]:
+        output = self.model_ft(image)
+        if label is None:
+            return {"pred": torch.argmax(output)}
+        loss = self.loss_fn(output, label)
```
epwalsh, line 42: We should also calculate the accuracy - at least when not in "training" model (`self.training` is `False`) - and return that along with the loss.

epwalsh (changes requested):
Just a couple more requests

examples/finetune_resnet/config.jsonnet
```diff
@@ -0,0 +1,79 @@
...
+            type: "torch::train",
+            model: {
+                type: pretrained_model,
+                num_classes: num_classes,
+                feature_extract: true,
+                use_pretrained: true,
+            },
+            dataset_dict: {"type": "ref", "ref": "transform_data"},
+            train_dataloader: single_device_dataloader,
+            validation_split: null,
```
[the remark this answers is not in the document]
SabhyaC26: fixed

examples/finetune_resnet/components.py
SabhyaC26: added

examples/finetune_resnet/components.py
```diff
@@ -0,0 +1,138 @@
...
+class Prediction(Step):
+    FORMAT: Format = JsonFormat()
+
+    def run(self, image_url: str, input_size: int, model: models) -> Dict[str, Any]:  # type: ignore
+        # download and store image
+        image_path = cached_path(image_url)
+        transformed_image = pil_loader(image_path, input_size, transform_type="test")
+
+        # pass image through transform
+        transformed_image = transformed_image.unsqueeze(0)
```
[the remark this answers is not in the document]
SabhyaC26: added

examples/finetune_resnet/resnet_steps.py
```diff
@@ -0,0 +1,139 @@
...
+        return "cat"
+    else:
+        return "dog"
+
+
+@Step.register("prediction")
+class Prediction(Step):
+    FORMAT: Format = JsonFormat()
+
+    def run(self, image_url: str, input_size: int, model: models, device: Optional[str] = "cpu") -> Dict[str, Any]:  # type: ignore
```
epwalsh, line 128:
```suggestion
    def run(self, image_url: str, input_size: int, model: Model, device: Optional[str] = "cpu") -> Dict[str, Any]:  # type: ignore
        model = model.to(device)
```

epwalsh (approved):
LGTM!

PR #153, opened 2022-01-23
Makes `Lazy` work with object that don't derive from `FromParams`

### LINKED ISSUE #146: `Lazy` should work properly with classes that don't derive from `FromParams`
When you use `Lazy(MyClass)`, it will try to call `MyClass.from_params`. It should do that only with `FromParams` classes. Otherwise, call `__init__`.

For now, the workaround is to write `Lazy(MyClass.__init__)`.

dirkgr:
Fixes #146.

### DIFF
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -203,6 +203,12 @@ class TorchTrainStep(Step):
             if amp:
                 raise ConfigurationError("AMP requires training on CUDA devices")
 
+        if validate_every is not None and validation_split is None:
+            raise ConfigurationError(
+                "You chave set a validation interval, but no invalidation split. "
+                "That's probably unintentional."
+            )
+
         is_distributed = False
         num_workers = 1
         if devices and len(devices) > 1:
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -4,6 +4,7 @@ import tempfile
 from itertools import islice
 from typing import Any, Dict, List, Optional, Set, cast
 
+import more_itertools
 import torch
 import torch.distributed as dist
 import torch.nn as nn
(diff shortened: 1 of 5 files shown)


tango/integrations/torch/train.py
```diff
@@ -203,6 +203,12 @@ def run(  # type: ignore[override]
             if amp:
                 raise ConfigurationError("AMP requires training on CUDA devices")
 
+        if validate_every is not None and validation_split is None:
```
epwalsh, line 206: I think you need to change the type of `validate_every` to `Optional[int]` and default to `None`.

epwalsh (commented):
Other than the one comment LGTM

dirkgr:
I fixed this. The way it is now, it will only validate at the end of the training job, or when `validate_every` is set. Don't we, by default, want to validate at the end of very epoch?

epwalsh:
> Don't we, by default, want to validate at the end of very epoch?

That sounds reasonable

dirkgr:
@epwalsh, I think this is ready for another review.

tango/integrations/torch/train.py
```diff
@@ -535,13 +543,14 @@ def save_state(step: int):
...
 
-    train_batch_iterator = Tqdm.tqdm(
+    train_batch_iterator_tqdm = Tqdm.tqdm(
         training_batches,
         desc="Training",
         initial=start_step,
         total=config.train_steps,
         disable=not config.is_local_main_process,
     )
+    train_batch_iterator = more_itertools.peekable(train_batch_iterator_tqdm)
```
epwalsh, line 553: What's with this?

dirkgr: I need to be able to look one batch ahead to find out whether this is the last batch of the epoch.

epwalsh: Oh, nevermind, I see

tango/integrations/torch/train_config.py
```diff
@@ -184,12 +184,6 @@ def should_checkpoint_this_step(self, step: int) -> bool:
         assert self.train_steps is not None
         return ((step + 1) % self.checkpoint_every == 0) or step == self.train_steps - 1
 
-    def should_validate_this_step(self, step: int) -> bool:
```
dirkgr, line 187: It's kind of shitty that all the `should_*` things are here, except for `should_validate`. But we don't have all the information we need here. We could leave the function here, and pass in the information of whether this is the last step of the epoch. It's just a little silly to say "if this parameter is `True` return `True` immediately and don't do anything."? But maybe that documents itself better.

epwalsh (approved):
LGTM

PR #155, opened 2022-01-24
set 'resume' to 'allow' instead of 'auto' when resuming W&B callback

epwalsh:


dirkgr (approved):
That's all it takes?

epwalsh:
> That's all it takes?

Resuming still worked before with "auto", but this should make it more robust.

PR #159, opened 2022-01-26
Add a required flag to the README for gpt2-example

schmmd:


README.md
```diff
@@ -90,7 +90,7 @@ cd tango
 Then run
 
 ```bash
-pip install -e .
+pip install -e .[all]
```
schmmd, line 93: I think most people installing from source will want this, but I can revert this change if you disagree.

examples/train_gpt2/README.md
```diff
@@ -7,5 +7,5 @@ otherwise Tango will fall back to CPU-only and it will be extremely slow.
 You can kick off a training job with the following command:
 
 ```
-$ tango run config.jsonnet
+$ tango run config.jsonnet -i components
```
schmmd, line 10: Sorry, I later discovered this too is needed in my previous PR.

README.md
epwalsh: I would leave this as-is since the next sentence below shows how to install with integrations, but I think it would be worth adding a note about the "all" integration.

schmmd: Do you disagree that most people would want `[all]` if they are installing from source?

epwalsh: No, but if we're going to change this line we should also update the lines below to explain how to install locally with only specific integrations or no integrations.

schmmd: I think I did exactly what you suggested already.  The next explanation is how to install a specific integration and then it explains how to install just the base library.

epwalsh: Oh, somehow I missed that. Sorry!

README.md
```diff
@@ -90,7 +90,7 @@ cd tango
...
 ```
 
 To install with a specific integration, such as `torch` for example, run
```
epwalsh, line 96:
```suggestion
To install with only a specific integration, such as `torch` for example, run
```

epwalsh (approved):
LGTM!

PR #160, opened 2022-01-27
Guess step dir better

dirkgr:
The impetus for this was that checking that a step dir is empty didn't work, because it already contained a work directory at the time. But a step dir with only a work directory still means the step is incomplete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The web UI now renders the step graph left-to-right.
 - The web UI now shows runs by date, with the most recent run at the top.
 - The web UI now shows steps in a color-coded way.
+- The `--include-package` flag now also accepts paths instead of module names.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The waiting message for `FileLock` is now clear about which file it's waiting for.
 - Added an easier way to get the default Tango global config
 - Plain old Python functions can now be used in `Lazy` objects.
+- `LocalWorkspace` now creates a symlink to the outputs of the latest run.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Tango is now better at guessing when a step has died and should be re-run.
+- Tango is now more lenient about registering the same class under the same name twice.
 
 
 ## [v0.4.0](https://github.com/allenai/tango/releases/tag/v0.4.0) - 2022-01-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- The waiting message for `FileLock` is now clear about which file it's waiting for.
+- Added an easier way to get the default Tango global config
+
 ### Fixed
 
 - Fixed bug where `Executor` would crash if `git` command could not be found.
+
 
 ## [v0.4.0](https://github.com/allenai/tango/releases/tag/v0.4.0) - 2022-01-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The waiting message for `FileLock` is now clear about which file it's waiting for.
 - Added an easier way to get the default Tango global config
+- Plain old Python functions can now be used in `Lazy` objects.
 
 ### Fixed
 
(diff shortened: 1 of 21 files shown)


tango/local_workspace.py
```diff
@@ -323,17 +352,21 @@ def find_or_add_step_info(step_or_unique_id: Union[Step, str]) -> StepInfo:
...
+                step_state_guess = self.guess_step_dir_state(step_dir) or step_info.state
+                if (
+                    step_state_guess == StepState.INCOMPLETE
+                    and step_info.state != StepState.INCOMPLETE
                 ):
-                    new_state = StepState.INCOMPLETE
-                if new_state != step_info.state:
                     step_info.start_time = None
                     step_info.end_time = None
                     d[unique_id] = step_info
```
epwalsh, line 362: Should we not update `step_info.state` to `INCOMPLETE` is this case?

dirkgr: `step_info.state` is a property that becomes `INCOMPLETE` when there is no start time and no end time.

CHANGELOG.md
```diff
@@ -20,11 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The web UI now shows runs by date, with the most recent run at the top.
 - The web UI now shows steps in a color-coded way.
 - The `--include-package` flag now also accepts paths instead of module names.
+- Tango is now better at guessing when a step has died and should be re-run.
```
epwalsh, line 23: This needs to be moved up to "Unreleased"

tango/local_workspace.py
```diff
@@ -275,9 +275,38 @@ def work_dir(self, step: Step) -> Path:
...
-    @staticmethod
-    def dir_is_empty(dir: Path):
-        return not any(True for _ in dir.iterdir())
+    @classmethod
+    def guess_step_dir_state(cls, dir: Path) -> Optional[StepState]:
+        """
+        Returns the status of a given step dir, to the best of our knowledge.
+
+        :param dir: the step dir to example
+        :return: our best guess at the state of the step, or ``None`` if we don't know
```
epwalsh, line 284: We're now using two different doc string styles in this repo. They both work, but I think it's better to be consistent.

tango/local_workspace.py
```diff
@@ -275,9 +275,38 @@ def work_dir(self, step: Step) -> Path:
...
+        # If the directory doesn't exist, the step is incomplete.
+        if not dir.exists():
+            return StepState.INCOMPLETE
+
+        # If the lock file exists and is locked, the step is running.
+        lock_file = dir / "lock"
+        if lock_file.exists():
+            lock = FileLock(lock_file)
+            try:
+                lock.acquire(0)
```
epwalsh, line 296: I think it would be a little safer to acquire with a small timeout, like 0.2 for example. That would help avoid the situation where two (or more) separate Python processes reach this line of code at the same time, both trying to acquire the lock immediately. Because only one will succeed, and the other(s) will erroneously think the step is running.

tango/local_workspace.py
```diff
@@ -275,9 +275,38 @@ def work_dir(self, step: Step) -> Path:
...
+                lock.release()
+            except TimeoutError:
+                return StepState.RUNNING
+
+        # If the directory is empty except for the work dir, the step is incomplete.
```
epwalsh, line 301: Shouldn't we ignore the lock file in addition to the work directory? I believe the lock file will probably exist for any incomplete step, since the lock file is created in the `step_starting()` method. As is, this check will fail to correctly identify a step as `INCOMPLETE` whenever the lock file exists.

epwalsh (changes requested):

CHANGELOG.md
dirkgr: dcdaad0293cdea55b143e27a79fca2e4a9d28934

tango/local_workspace.py
dirkgr: I don't see it. I'm using the documentation for `Step` as the gold standard, and you wrote that one, didn't you? What's the difference?

I like this style primarily because IntelliJ fills it in automatically, but there are probably ways to adjust how it does that.

dirkgr: Oh, `TorchTrainStep` is very different.

tango/local_workspace.py
dirkgr: 8d8785f

tango/local_workspace.py
dirkgr: I think you're right. 81d9276

dirkgr:
It turns out this only worked because it wasn't ignoring the lock file. Once it started ignoring the lock file, I had to actually fix it.

tango/local_workspace.py
```diff
@@ -277,9 +277,42 @@ def work_dir(self, step: Step) -> Path:
...
+
+        # If the lock file exists and is locked, the step is running.
+        lock_file = dir / "lock"
+        if lock_file.exists():
+            lock = FileLock(lock_file)
+            try:
+                lock.acquire(0.1)
+                lock.release()
+            except TimeoutError:
+                return {StepState.RUNNING}
```
dirkgr, line 301: @epwalsh, is this reliable? If the lock file isn't locked, can we safely assume that the step can't possibly be running?

I could swear I saw this fail on the Mac filesystem, but I can't reproduce any problems with it now.

epwalsh: I think that's safe to assume

epwalsh (approved):

PR #161, opened 2022-01-27
Some fixes for our favorite Tango component

dirkgr:
Changelog says it all.


dirkgr:
@ZhaofengWu, this is the error message we were talking about.

tango/common/from_params.py
```diff
@@ -141,7 +141,21 @@ def infer_method_params(
...
+            try:
+                parameters[param_name] = param.replace(
+                    annotation=get_type_hints(method)[param_name]
+                )
+            except TypeError as e:
+                if "'type' object is not subscriptable" in str(e):
+                    new_e = TypeError(
+                        f"Could not determine the type annotations for `{method.__qualname__}`. Did you use the "
+                        "lowercase `dict`, `list`, `set`, or `tuple` in the type annotations instead of the "
+                        "uppercase `Dict`, `List`, `Set`, `Tuple`?"
```
epwalsh, line 153: If I understand correctly, this is catching the case where the annotation is `dict` instead of, say, `Dict[str, Any]` or whatever, right? But `dict[str, Any]` is still okay, no? If so, I think we should make this error message more clear. Something like: "Did you forget to set the type's arguments in brackets? E.g. dict[str, Any]"

dirkgr: `dict[str, Any]` is OK in the sense that you can put anything there and Python will ignore it anyways. But as soon as you try to use it it fails.

epwalsh: > Python will ignore it anyways

What do you mean? Can `FromParams` not access the `str` and `Any` from the annotations?

> But as soon as you try to use it it fails.

Try to use it how? And how does it fail?

dirkgr: >> Python will ignore it anyways

> What do you mean? Can `FromParams` not access the `str` and `Any` from the annotations?

Python, by default, does not do anything with the annotations. `FromParams` does, and promptly fails when the annotation contains `dict`.

> > But as soon as you try to use it it fails.

> Try to use it how? And how does it fail?

When you try to use it in `FromParams`. In detail, the call that doesn't work is `typing.get_type_hints()`. I wanted to give you an example that you can paste into a Python console, but it turns out that the Python console attempts to parse the type hints. Who knew?

```
>>> from typing import *
>>> def f(d: dict[str, str]) -> None:
...     pass
...
  File "<stdin>", line 1, in <module>
TypeError: 'type' object is not subscriptable
```

That is the exact error you get in `FromParams` when it calls `get_type_hints()`.

dirkgr: I guess I can trigger the error by making the type hint a string:

```
>>> def f(d: 'dict[str, str]') -> None:
...     pass
...
>>> get_type_hints(f)
  File "<stdin>", line 1, in <module>
  File "/opt/miniconda3/envs/meta-learn-prompt/lib/python3.7/typing.py", line 1013, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "/opt/miniconda3/envs/meta-learn-prompt/lib/python3.7/typing.py", line 263, in _eval_type
    return t._evaluate(globalns, localns)
  File "/opt/miniconda3/envs/meta-learn-prompt/lib/python3.7/typing.py", line 467, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: 'type' object is not subscriptable
```

dirkgr: This is the line: https://github.com/allenai/tango/blob/main/tango/common/from_params.py#L144

epwalsh: You're running into these errors because you're using Python 3.7. `d: dict[str, str]` is not valid syntax in Python 3.7. It's only valid as of Python 3.10 or if you do `from __future__ import annotations`.

epwalsh: ** valid as of 3.9

epwalsh: We already have a test for this: https://github.com/allenai/tango/blob/5d602fbf83998020288a986e2ef1b356e1c113ea/tests/common/from_params_test.py#L1038-L1052

dirkgr: Should I change the error message to say "Please use `from __future__ import annotations`"?

epwalsh: That would be good. But only if the Python version is < 3.9

dirkgr: No, I checked this, that's not the case. When you don't use `from __future__ import annotations`, your whole file will fail to import. None of our code ever runs.

When you do use `from __future__ import annotations`, the file will import, but `param.annotation` is a `str`, which needs parsing. That's why we're in this branch in the first place. `get_type_hints()` does that parsing, but regardless of `annotations`, it will fail if your annotation is `dict`.

In Python >= 3.10, we will never even go into this branch.

The comment right above that part of the code is actually quite accurate and informative. Someone very smart must have written that.

epwalsh: Okay, so correct me if I'm wrong:

First of all, none of this is an issue with Python 3.9 or newer. With Python 3.7 or 3.8 and `from __future__ import annotations`, a type hint like `d: dict` is technically okay (but not very informative), while a type hint like `d: dict[str, str]` will fail with "TypeError: 'type' object is not subscriptable."

In light of that, this makes sense to me now. Before, I was assuming that a type hint like `d: dict[str, str]` would still be okay when `__annotations__` was imported. That's not the case.

I think we can still make the error message a little more helpful. For one, it's not just `dict`, `list`, `set`, and `tuple` that this is an issue for, but really any type. Though, most likely, it will probably come from one of the many standard collections listed [here](https://www.python.org/dev/peps/pep-0585/).

See my other comment for a suggestion of how to improve it.

tango/common/from_params.py
```diff
@@ -141,7 +141,21 @@ def infer_method_params(
...
+                    )
```
epwalsh, line 154: How about:
```suggestion
                if "'type' object is not subscriptable" in str(e):
                    # This can happen when someone uses a type hint like `dict[str, str]`
                    # instead of `Dict[str, str]`.
                    err_msg = (
                        f"Failed to parse the type annotation `{param.annotation}` "
                        f"from `{cls.__qualname__}.{method.__name__}()`."
                    )

                    if "[" in param.annotation:
                        # Check if there is an equivalent generic in the `typing` module.
                        import typing

                        type_, *_ = param.annotation.split("[", 1)
                        for possible_typing_equivalent in {type_, type_.title()}:
                            if hasattr(typing, possible_typing_equivalent):
                                err_msg += (
                                    f" Try using `{possible_typing_equivalent}` "
                                    "from the `typing` module instead."
                                )
                                break

                    new_e = TypeError(err_msg)
                    new_e.__cause__ = e
```

This would result in an error message like:

```
TypeError: Failed to parse the type annotation `dict[str, str]` from `Foo.__init__()`. Try using `Dict` from the `typing` module instead.
```

dirkgr: I like it.

dirkgr: What's the difference between `{cls.__qualname__}.{method.__name__}` and `{method.__qualname__}`?

epwalsh: `method.__qualname__` will look like unrelated garbage for dataclasses, unfortunately. E.g. `__create_fn__.<locals>.__init__`.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- When you use `dict` instead of `Dict` in your type annotations, you now get a legible error message. Same for `List`, `Tuple`, and `Set`.
```
epwalsh, line 32: This ended up in the v0.4.0 section instead of Unreleased

CHANGELOG.md
```diff
@@ -38,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure tqdm log lines always make it into the log file `out.log` even when log level is `WARNING` or `ERROR`.
+- When you register the same name twice using `Registrable`, you get an error message. That error message now contains the correct class name.
```
epwalsh, line 42: Same here

epwalsh (approved):

PR #167, opened 2022-01-31
use virtualenv within Docker

epwalsh:
Changes the Dockerfile to use a Python virtual environment so we don't mess with the system Python packages on the image. I think this is good practice in general. See https://pythonspeed.com/articles/activate-virtualenv-dockerfile/ for an explanation of how the virtual environment is activated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,4 +47,4 @@ RUN pip install --no-cache-dir --no-deps .[all]
 WORKDIR /app/
 RUN rm -rf /stage/
 
-ENTRYPOINT ["/usr/local/bin/tango"]
+ENTRYPOINT ["/opt/venv/bin/tango"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,10 +18,15 @@ RUN apt-get update && apt-get install -y \
     python3-dev \
     python3-pip \
     python3-setuptools \
+    python3-venv \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip install --no-cache-dir --upgrade pip setuptools wheel
+# Create and activate a virtual environment so we don't mess with system packages.
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv /opt/venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN pip install --upgrade pip setuptools wheel
 
 # Install torch ecosystem before everything else since this is the most time-consuming.
 # If you need a different version of CUDA, change this instruction according to the official


dirkgr (commented):
Why virtualenv instead of conda, when we all use conda for development?

epwalsh:
> Why virtualenv instead of conda, when we all use conda for development?

`venv` is more direct and lightweight. What benefit does conda have?

(p.s. I don't use conda for development)

dirkgr:
Conda works for non-Python dependencies as well, so it can cover more of the stack, be less dependent on the system that it runs on. They actually test the dependencies together. And it's the first-among-equals release method for PyTorch.

epwalsh:
> Conda works for non-Python dependencies as well

What non-Python dependencies do we need? Not the CUDA toolkit, because our base image already comes with that. Using conda to install torch would result in a duplicate CUDA toolkit on the image.

> be less dependent on the system that it runs on.

I don't think that's a concern for a Docker image, where we completely control the environment.

dirkgr:
>> Conda works for non-Python dependencies as well

> What non-Python dependencies do we need?

This is not a perfect list since I also install `ipython`, and the way I filter makes it include torch itself, but it's still an impressive list:

```
dirkg@ip-192-168-1-194 ~/D/tango (DocumentationRefresh)> conda list | fgrep -v pypi                                                             (tango)
# packages in environment at /opt/miniconda3/envs/tango:
#
# Name                    Version                   Build  Channel
ai2-tango                 0.4.0rc5                  dev_0    <develop>
appnope                   0.1.2           py38hecd8cb5_1001
backcall                  0.2.0              pyhd3eb1b0_0
blas                      1.0                         mkl
bzip2                     1.0.8                h1de35cc_0
ca-certificates           2021.10.26           hecd8cb5_2
certifi                   2021.10.8        py38hecd8cb5_0
decorator                 5.1.0              pyhd3eb1b0_0
ffmpeg                    4.3                  h0a44026_0    pytorch
freetype                  2.11.0               hd8bbffd_0
gettext                   0.21.0               h7535e17_0
giflib                    5.2.1                haf1e3a3_0
gmp                       6.2.1                h23ab428_2
gnutls                    3.6.15               hed9c0bf_0
icu                       58.2                 h0a44026_3
intel-openmp              2021.4.0          hecd8cb5_3538
ipython                   7.29.0           py38h01d92e1_0
jedi                      0.18.0           py38hecd8cb5_1
jpeg                      9d                   h9ed2024_0
lame                      3.100                h1de35cc_0
lcms2                     2.12                 hf1fd2bf_0
libcxx                    12.0.0               h2f01273_0
libffi                    3.3                  hb1e8313_2
libiconv                  1.16                 h1de35cc_0
libidn2                   2.3.2                h9ed2024_0
libpng                    1.6.37               ha441bb4_0
libtasn1                  4.16.0               h9ed2024_0
libtiff                   4.2.0                h87d7836_0
libunistring              0.9.10               h9ed2024_0
libuv                     1.40.0               haf1e3a3_0
libwebp                   1.2.0                hacca55c_0
libwebp-base              1.2.0                h9ed2024_0
libxml2                   2.9.12               hcdb78fc_0
llvm-openmp               12.0.0               h0dcd299_1
lz4-c                     1.9.3                h23ab428_1
matplotlib-inline         0.1.2              pyhd3eb1b0_2
mkl                       2021.4.0           hecd8cb5_637
mkl-service               2.4.0            py38h9ed2024_0
mkl_fft                   1.3.1            py38h4ab4a9b_0
mkl_random                1.2.2            py38hb2f4e1b_0
ncurses                   6.3                  hca72f7f_1
nettle                    3.7.3                h230ac6f_1
numpy                     1.21.2           py38h4b4dc7a_0
numpy-base                1.21.2           py38he0bd621_0
olefile                   0.46               pyhd3eb1b0_0
openh264                  2.1.0                hd9629dc_0
openssl                   1.1.1l               h9ed2024_0
parso                     0.8.2              pyhd3eb1b0_0
pexpect                   4.8.0              pyhd3eb1b0_3
pickleshare               0.7.5           pyhd3eb1b0_1003
pillow                    8.4.0            py38h98e4679_0
pip                       21.2.4           py38hecd8cb5_0
prompt-toolkit            3.0.20             pyhd3eb1b0_0
ptyprocess                0.7.0              pyhd3eb1b0_2
pygments                  2.10.0             pyhd3eb1b0_0
python                    3.8.12               h88f2d9e_0
pytorch                   1.10.0                  py3.8_0    pytorch
readline                  8.1                  h9ed2024_0
setuptools                58.0.4           py38hecd8cb5_0
six                       1.16.0             pyhd3eb1b0_0
sqlite                    3.36.0               hce871da_0
tk                        8.6.11               h7bc2e8c_0
torchaudio                0.10.0                 py38_cpu    pytorch
torchvision               0.11.1                 py38_cpu    pytorch
traitlets                 5.1.0              pyhd3eb1b0_0
typing_extensions         3.10.0.2           pyh06a4308_0
wcwidth                   0.2.5              pyhd3eb1b0_0
wheel                     0.37.0             pyhd3eb1b0_1
xz                        5.2.5                h1de35cc_0
zlib                      1.2.11               h1de35cc_3
zstd                      1.4.9                h322a384_0
```

> Not the CUDA toolkit, because our base image already comes with that. Using conda to install torch would result in a duplicate CUDA toolkit on the image.

Well, I never understood why we have to manage CUDA ourselves. We had that discussion once before and left it unresolved.

>> be less dependent on the system that it runs on.

> I don't think that's a concern for a Docker image, where we completely control the environment.

Of course we have more flexibility when we determine ourselves which package versions run on top of which library versions. But there are people out there that already do this packaging for us. The conda people are among them. What's the value-add we're providing?

epwalsh:
Most of items in that list are Python packages... What's on that list that we would need to install separately on the Docker image? AFAIK PyTorch wheels come with everything they need, otherwise conda would really be the only practical way to install PyTorch.

> Well, I never understood why we have to manage CUDA ourselves.

I wouldn't say we're managing CUDA ourselves. That's the last thing I want. We're using an official NVIDIA CUDA image as our base image to _avoid_ managing CUDA ourselves.

To be clear, I'm not totally against using conda in Docker images. But if we do go that route, I don't think it makes sense to start from the NVIDIA CUDA images because conda will just install another CUDA toolkit. And I think we can both agree that we don't want duplicate CUDA toolkits on our images. So that means we should start from a different base image, but that worries me because we've ran into issues before on the AllenNLP images when we weren't using NVIDIA CUDA base images.

dirkgr:
There is a bunch of stuff on that list where it relies on the system libraries instead. `openssl`, `libz`, `tk`, `readline`, `ncurses`, all the image loading libraries, `intel-openmp` (which will probably be an install-time if not runtime dependency for other Python libraries that use that style of multiprocessing). I don't know how `mkl` gets installed when you `pip install torch`. Maybe you just don't get `mkl` and that's your problem?

When you use this stuff from the system, `error while loading shared libraries: libfoo.so: cannot open shared object file: No such file or directory` always lurks.

If we rely on CUDA from conda, we don't have to ship specific images anymore for each CUDA version we want to be compatible with. But I don't know about all the edge cases.

epwalsh:
I'm not sure about `mkl` and some of the others, but I do know that all of the image loading libraries come with the wheel of `torchvision`.

> If we rely on CUDA from conda, we don't have to ship specific images anymore for each CUDA version we want to be compatible with.

That's not true. The situation doesn't change. Either way we still ship images that are tied to a specific version of the CUDA toolkit, because **the images will have a specific version of the CUDA toolkit installed on them**. The CUDA toolkit either comes with the base image (the current case) or is installed from conda (what you're proposing we do).

epwalsh:
>> If we rely on CUDA from conda, we don't have to ship specific images anymore for each CUDA version we want to be compatible with.

> That's not true. The situation doesn't change. Either way we still ship images that are tied to a specific version of the CUDA toolkit, because the images will have a specific version of the CUDA toolkit installed on them. The CUDA toolkit either comes with the base image (the current case) or is installed from conda (what you're proposing we do).

Also the fact that PyTorch ships images for specific CUDA versions is a pretty good indication that it's important. https://hub.docker.com/r/pytorch/pytorch/tags

epwalsh:
Anyway, for the sake of moving things forward and because I really don't feel strongly about conda, I propose we do the following:
1. We merge this PR as-is assuming we can agree that it improves the existing Docker image. This gives us something more reliable to roll back to later if we find issues with the conda approach.
2. In a separate PR I'll change the Dockerfile to use conda for installing all of our Python AND non-Python dependencies. And I'll also move to a non-NVIDIA base image so that we don't have duplicate CUDA toolkits on the image. This new Dockerfile will probably look similar to [`pytorch / Dockerfile`](https://github.com/pytorch/pytorch/blob/master/Dockerfile).

dirkgr:
I'm fine with shipping this, but I want to get to the bottom of this at some point, both so we make the right decision for our Docker images, and for my own understanding of the matter.

The way I see it, these Docker images are part of a layer cake. The layer underneath us has the NVIDIA drivers. The layer above us expects PyTorch to work. Everything in between is up to us. Since CUDA 11, it does not matter anymore what the exact CUDA version is, as long as the driver version is at least 450.80.02 (see https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-major-component-versions, Table 2). Since the exact CUDA version doesn't matter, we might as well let conda decide what it's going to be, and stay out of it ourselves. Is any part of this picture wrong?

One issue I have had in the past with CUDA from conda is that some versions were not compiled with the right compute capabilities for the GPU I had at the time. I did not rigorously track this down. But before we commit 100% to conda, someone (probably me) should do that and confirm that we'll be fine.

epwalsh:
> Since the exact CUDA version doesn't matter, we might as well let conda decide what it's going to be, and stay out of it ourselves.

Whether using conda or pip, we still have to decide which CUDA version to use with the same degree of specificity when we install PyTorch on the image. This is apparent from the official PyTorch install instructions (https://pytorch.org/). In particular, you have to choose CUDA 10.2, CUDA 11.3, or no CUDA. This is why I'm saying we still have to make that decision, and therefor we still have to ship images that are tied to a specific version of the CUDA toolkit.

epwalsh:
I'm happy to look into this further, I'll have to come back to our Docker images anyway for Beaker execution.

dirkgr:
PyTorch offers these versions because they don't know how you want to run. They don't package it together with CUDA, so they have to have multiple versions available. It's different for us (and conda). We ship both PyTorch and CUDA in one package, so we can decide ourselves. Or let conda do it.

epwalsh:
> They don't package it together with CUDA, so they have to have multiple versions available. 

I don't think this is accurate. Here is a quote within the past year from PyTorch maintainer:

> Since [conda binaries and pip wheels] ship with their own CUDA runtime, you would only need a local NVIDIA driver corresponding to the CUDA runtime you are selecting.
> https://discuss.pytorch.org/t/is-it-required-to-set-up-cuda-on-pc-before-installing-cuda-enabled-pytorch/60181/15

> It's different for us (and conda). We ship both PyTorch and CUDA in one package, so we can decide ourselves. Or let conda do it.

When you install torch with conda, you **still** have to choose the CUDA version. Conda won't decide on it's own.

So, in terms of the CUDA toolkit, it seems like it doesn't matter whether we install torch on our images via pip wheels or conda. Either way we get the CUDA toolkit. And that means that starting from the nvidia base images results in duplicate versions of the CUDA toolkit on the image. So we shouldn't start from nvidia base images.

dirkgr:
Like I said, I'm have to merge this as is. I just want to get to the bottom of this at some point. It doesn't have to be in this thread.

I did not know that pip wheels also come with CUDA inside. So I can `pip install torch` on any box, and as long as the driver is there I'm good to go? I thought pip wheels link against CUDA libraries that need to be present, but if this is not the case, that makes things easier.

>  It's important that we tag our images with the CUDA version that PyTorch was built with to ensure people will only use those images if their drivers are compatible.

If all drivers since 450.80.02 are compatible, is this really still important? 450.80.02 came out in July-ish last year.

epwalsh:
> So I can pip install torch on any box, and as long as the driver is there I'm good to go?

Yes, that sounds like the case, at least for recent versions of PyTorch (within the last year).

> If all drivers since 450.80.02 are compatible, is this really still important? 450.80.02 came out in July-ish last year.

I think it's still worth doing. I'm sure there are plenty of researchers out there who don't have an infrastructure team as good as ours that keeps their drivers up-to-date. Even if that's not a reasonable assumption, I think we benefit from mirroring what PyTorch does, because that's what people will expect.

dirkgr:
Here is a case I didn't think about earlier: If you want to use DeepSpeed's `FusedAdam`, you have to have `nvcc` of the right version installed. Conda does not have this, but conda-forge has my back with `cudatoolkit-dev`.

PR #171, opened 2022-02-02
Lenient registration

dirkgr:
I think the comments in the code explain pretty well what's going on here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Tango is noew more lenient about registering the same class under the same name twice.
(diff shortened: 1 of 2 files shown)


dirkgr:
This problem is causing a right mess though, and this is not the only scenario in which it fails.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed bug where `Executor` would crash if `git` command could not be found.
 
+### Changed
+
+- Tango is noew more lenient about registering the same class under the same name twice.
```
epwalsh, line 21:
```suggestion
- Tango is now more lenient about registering the same class under the same name twice.
```

epwalsh (approved):

PR #172, opened 2022-02-04
use conda in Docker image, multi-stage build

epwalsh:
Adapted from PyTorch's Dockerfile: https://github.com/pytorch/pytorch/blob/master/Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,14 +19,14 @@ RUN curl -fsSL -v -o ~/miniconda.sh -O  https://repo.anaconda.com/miniconda/Mini
     chmod +x ~/miniconda.sh && \
     ~/miniconda.sh -b -p /opt/conda && \
     rm ~/miniconda.sh && \
-    /opt/conda/bin/conda install -y python=${PYTHON_VERSION} ipython && \
+    /opt/conda/bin/conda install -y "python=${PYTHON_VERSION}" ipython && \
     /opt/conda/bin/conda clean -ya
 
 FROM conda as conda-installs
 ARG PYTHON_VERSION=3.9
 ARG CUDA_VERSION=11.3
 ARG INSTALL_CHANNEL=pytorch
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,50 +1,60 @@
...
+    ~/miniconda.sh -b -p /opt/conda && \
+    rm ~/miniconda.sh && \
+    /opt/conda/bin/conda install -y "python=${PYTHON_VERSION}" ipython && \
+    /opt/conda/bin/conda clean -ya
+
+FROM conda as conda-installs
+ARG PYTHON_VERSION=3.9
+ARG CUDA_VERSION=11.3
+ARG INSTALL_CHANNEL=pytorch
+RUN /opt/conda/bin/conda install -c "${INSTALL_CHANNEL}" -y "python=${PYTHON_VERSION}" pytorch torchvision torchtext "cudatoolkit=${CUDA_VERSION}" && \
```
dirkgr, line 29: Can we also install `cudatoolkit-dev=${CUDA_VERSION}` from `conda-forge`, and `mpi4py` from regular conda?

You need those things for Deepspeed, or other libraries that ship fused GPU operations.

dirkgr: Honestly I don't know why `cudatoolkit` doesn't just ship with the dev tools integrated. I think when you install CUDA from Nvidia's packages they all come together, so the conda people went out of their way to un-bundle it.

dirkgr (changes requested):

Dockerfile
epwalsh: Done in https://github.com/allenai/docker-images/pull/28/commits/9eedd8eeb25c9c89ccecd4ceee4765459e09b44d

epwalsh: I've moved the base over to our common docker images in https://github.com/allenai/docker-images/pull/28 so that other projects can benefit from this.

epwalsh:
Ugh, these Docker builds are sooo slow

dirkgr (approved):

PR #173, opened 2022-02-04
Make the optimizers from the transformers package available

dirkgr:


dirkgr:
@epwalsh has a better version of this.

PR #175, opened 2022-02-07
make 'resolve_module_name()' tests self-contained

epwalsh:


dirkgr (approved):

h-vetinari:
Thanks a lot for the quick response @epwalsh!

Note that this change itself does not yet make the tests self-contained (in the sense of being run out of-tree), but it's enough that it'll work with the existing patch in conda-forge for `TangoTestCase` (very similar to https://github.com/allenai/allennlp/pull/5560, which I'd like to use as a stand-in for discussing potential upstreaming).

Issue #182, opened 2022-02-09
Use Beaker to run GPU tests


epwalsh:
## Tasks
- [x] Set up a scalable, preemptible,  cheap cloud cluster on Beaker: https://beaker.org/cl/ai2/tango-gpu-tests/details
- [x] Set up a Beaker workspace: https://beaker.org/ws/ai2/tango-testing
- [ ] #183
- [x] #184 (merged into 183)

PR #183, opened 2022-02-09
Add Docker image for running GPU tests on Beaker

epwalsh:
Part 1 of #182

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,7 +7,6 @@
 .mypy_cache
 docs
 examples
-scripts
 tests
 test_fixtures
 dist
--- a/.github/workflows/docker_testing.yml
+++ b/.github/workflows/docker_testing.yml
@@ -1,0 +1,71 @@
+name: Docker testing
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'Dockerfile.test'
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   transformers library and now registered under the corresponding class in the torch integration.
   For example, transformers `Adafactor` optimizer is registered as an `Optimizer` under the name
   "transformers::Adafactor". More details can be found in the documentation for the transformers integration.
+- Added a '@requires_gpus' decorator for marking tests as needing GPUs. Tests marked with this will be ran in the "GPU Tests" workflow
```
AkshitaB, line 19:
```suggestion
- Added a '@requires_gpus' decorator for marking tests as needing GPUs. Tests marked with this will be run in the "GPU Tests" workflow
```

epwalsh: 🤦‍♂️

dirkgr (commented):
External contributors can still see the logs from failed tests, right?

epwalsh:
> External contributors can still see the logs from failed tests, right?

Currently this workflow relies on access to the `BEAKER_TOKEN` secret, which won't be available to external contributors unless one of us triggers their workflow. That's not ideal, but at least it's easy for us to trigger their workflows. For example, when one of us re-runs a failed workflow, that counts as us triggering it, so it will then have access to the secrets.

PR #192, opened 2022-02-11
Close sqlite harder

dirkgr:
Sometimes Tango hangs after finishing. This may or may not be a fix. I can't be sure, because I can't repro.


dirkgr:
@h-vetinary confirms that this fixed a hang in his tests. It's unclear why it doesn't hang the same way when we run it.

AkshitaB (approved):

PR #193, opened 2022-02-12
Step graph from file

dirkgr:
In the https://github.com/allenai/lm-robustness project, I want to parse a Jsonnet file and then interact with the step graph programmatically. This makes it easy.


AkshitaB (changes requested):
Looks good. Will be helpful to have a test case.

dirkgr:
Right you are. [c5c2490](https://github.com/allenai/tango/pull/193/commits/c5c2490d73bd8df651ba36433dd25476ce7608d2)

dirkgr:
I found two different problems while writing that test. We are very dependent on the order the tests run in. This is not ideal.

AkshitaB (approved):

PR #195, opened 2022-02-14
Rename "ExecutorMetadata" -> "StepExecutionMetadata"

epwalsh:
This just makes more sense now that this class is decoupled from the `Executor`.


dirkgr (approved):
Gotta get those API-breaking renames in before version 1.0!

PR #200, opened 2022-02-17
Show uncacheable steps in UI

AkshitaB:
Fixes #

- Uncacheable steps now visible in UI

Screen Shot 2022-02-16 at 4 16 11 PM

## Before submitting

## After submitting

### DIFF
--- a/tango/local_workspace.py
+++ b/tango/local_workspace.py
@@ -560,8 +560,8 @@ class LocalWorkspace(Workspace):
     def _run_step_info_file(self, name: str):
         return self.runs_dir / name / "stepinfo.json"
 
-    def _save_registered_run(self, name: str, all_steps: Iterable[Step]):
-        unique_ids = []
+    def _save_registered_run(self, name: str, all_steps: Iterable[Step]) -> None:
+        step_unique_ids = {}
         with SqliteDict(self.step_info_file) as d:
             for step in all_steps:
                 try:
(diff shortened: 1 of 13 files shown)


tango/local_workspace.py
```diff
@@ -576,12 +570,7 @@ def registered_run(self, name: str) -> Run:
             raise KeyError(name)
         with SqliteDict(self.step_info_file, flag="r") as d:
             steps_for_run = {}
-            for step_symlink in run_dir.iterdir():
-                if not step_symlink.is_symlink():
-                    continue
-                step_name = str(step_symlink.name)
-                unique_id = str(step_symlink.resolve().name)
-                step_info = d[unique_id]
+            for step_name, step_info in d.items():
```
dirkgr, line 573: This will run through all steps in sqlite. The old code only ran over all steps that are part of that run.

AkshitaB: So, should we just create empty links for uncacheable steps?

tango/local_workspace.py
```diff
@@ -574,18 +554,59 @@ def registered_run(self, name: str) -> Run:
...
+                except KeyError:
+                    d[step.unique_id] = StepInfo(
+                        step.unique_id,
+                        step.name if step.name != step.unique_id else None,
+                        step.__class__.__name__,
+                        step.VERSION,
+                        {dep.unique_id for dep in step.dependencies},
+                        step.cache_results,
+                    )
+                d.commit()
```
dirkgr, line 580: Don't `commit()` every time. Just do it after all steps are done. I realize the old code did this, but fewer bigger transactions are going to be faster than lots of small ones.

tango/local_workspace.py
```diff
@@ -574,18 +554,59 @@ def registered_run(self, name: str) -> Run:
...
+                unique_ids.append(step.unique_id)
+
+            run_step_info_file = self._run_step_info_file(name)
+            with open(run_step_info_file, "w") as file_ref:
+                json.dump(unique_ids, file_ref)
+
+    def _load_registered_run(self, name: str):
```
dirkgr, line 587: Return type?

tango/local_workspace.py
```diff
@@ -574,18 +554,59 @@ def registered_run(self, name: str) -> Run:
...
             raise KeyError(name)
-        with SqliteDict(self.step_info_file, flag="r") as d:
-            steps_for_run = {}
+        steps_for_run = self._load_registered_run(name)
+        return Run(name, steps_for_run, datetime.fromtimestamp(run_dir.stat().st_ctime))
+
+    def _run_step_info_file(self, name: str):
+        return self.runs_dir / name / "stepinfo.json"
+
+    def _save_registered_run(self, name: str, all_steps: Iterable[Step]):
```
dirkgr, line 563: Return type?

dirkgr (changes requested):
The same step can have a different name in different runs. For one experiment, a model might be the "final_model", and in another it might be the "baseline_model". It's bad design, my fault, that the `Step` and `StepInfo` classes even have a `step_name` member.

Because of this it is not safe to read the `step_name` out of the step info classes. If a different run ran in the meantime, a step's `step_name` could have changed. The step names have to stay with the run. It used to be the name of the symlink pointing to the result directory. Now, I recommend you make the json file containing the steps be a mapping, mapping step names to unique ids, to preserve this property.

PR #201, opened 2022-02-18
Changes to allow support for more Workspace and StepCache implementations

epwalsh:
This is the first pull request for the [Tango remote execution and workspaces](https://docs.google.com/document/d/1HkZW32Gn9AocMIGpNvUMp3qus8aJjW7k8Nt1F_5AXFM/edit#heading=h.83mquxrqsg59) project.

This PR contains minor API and internal module organization changes to prepare for the addition of remote workspaces. I've tried hard to keep the commit history clean to make reviewing this PR easier, so I'd recommend reviewing this one commit at a time:
- [Allow Workspace configuration from 'tango run'](https://github.com/allenai/tango/pull/201/commits/5237adcd58f764f0357ec9d11dc1216a22e71a04)
  - Basically this makes it so users can actually use different `Workspace` implementations from `tango run`.
  - It adds a `Workspace.from_url()` method and a command-line option (for `tango run`) called `-w/--workspace` that takes a workspace url like "local:///tmp/workspace".
  - Also lets users specify a default workspace in their `tango.yml` configuration file so they don't always have to pass `-d/--workspace-dir` or `-w/--workspace`.
- [move Workspace implementations to their own submodule 'tango.workspaces'](https://github.com/allenai/tango/pull/201/commits/421f6a2f7935bf5e90942ea265e3c9a5fea66d06)
  - This is simply an update to internal module organization. Previously we had the modules `tango.workspace`, which included the base class and also the `MemoryWorkspace` implementation, and then `tango.local_workspace`, which had the `LocalWorkspace` implementation. This "flat" module structure is not consisted with how we usually organize things and doesn't "scale" well to adding new `Workspace` implementations. So I created a new submodule `tango.workspaces` and put both implementations (`MemoryWorkspace` and `LocalWorkspace`) in there.
  - I also moved the `StepExecutionMetadata` (and related classes) to `tango.workspace`, since I think this will be used by other `Workspace` implementations in the future, not just `LocalWorkspace`.
- [move StepCache implementations to their own submodule 'tango.step_caches'](https://github.com/allenai/tango/pull/201/commits/715475fb059e6a738c8b58828ae91d308561d0c2)
  - This is just another update to internal module organization, similar to the above, but for `StepCache` and its implementations. It creates a new submodule `tango.step_caches` and puts `MemoryStepCache` and `LocalStepCache` in there.
 - [deprecate -d/--workspace-dir, allow -w to take a path](https://github.com/allenai/tango/pull/201/commits/2357eae3615ba4e6ce8c080ee3b4c2dfde5d9e17)
   - The `-d/--workspace-dir` option is now deprecated. When someone uses that option, we issue a deprecation warning with a suggestion to use `-w/--workspace` instead.
   - The `-w/--workspace` option also take a raw path, like `/tmp/workspace`, which is equivalent to `local:///tmp/workspace`. This is for convenience so users can benefit from path completion on the command line.

### DIFF
--- a/tango/workspace.py
+++ b/tango/workspace.py
@@ -223,7 +223,7 @@ class Workspace(Registrable):
         would give you a :class:`~tango.workspaces.LocalWorkspace` in the directory ``/tmp/workspace``.
         """
         parsed = urlparse(url)
-        workspace_type = parsed.scheme
+        workspace_type = parsed.scheme or "local"
         workspace_cls = cast(Workspace, cls.by_name(workspace_type))
         return workspace_cls.from_parsed_url(parsed)
 
--- a/tango/workspace.py
+++ b/tango/workspace.py
@@ -1,22 +1,28 @@
-import copy
+import getpass
+import logging
+import os
+import platform
+import socket
+import sys
+import time
 from abc import abstractclassmethod, abstractmethod
 from contextlib import contextmanager
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from datetime import datetime, timedelta
 from enum import Enum
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import Dict, Iterable, Iterator, List, Optional, Set, TypeVar, Union, cast
+from typing import Any, Dict, Iterable, List, Optional, Set, TypeVar, Union, cast
 from urllib.parse import ParseResult, urlparse
 
 import dill
-import petname
 
-from tango import step_cache
-from tango.common import Registrable
-from tango.common.util import exception_to_string
+from tango.common import FromParams, Registrable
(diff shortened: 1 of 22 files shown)


tango/workspace.py
```diff
@@ -306,115 +334,212 @@ def registered_run(self, name: str) -> Run:
         """
         raise NotImplementedError()
 
+    @contextmanager
+    def capture_logs_for_run(self, name: str):
+        """
+        Should return a context manager that can be used to capture the logs for a run.
+
+        By default, this doesn't do anything.
```
dirkgr, line 342: Can you add an example?

dirkgr (approved):

AkshitaB (approved):

PR #204, opened 2022-02-19
Multicore Parallelism

AkshitaB:

- Start with `__main__` and `Executor`: Allows a single step to be executed on its own (this is done in the main process).
- Changes to `StepGraph` and `Format` that allow converting the objects `to_config`.
- `MulticoreExecutor`: Creates individual processes for every step run. Start with `execute_step_graph`.

Todo:
- [x] Better logging. Currently, the main out.log has info for when steps are spawned into their own processes, but logs from individual steps are in their own out.log files. We may want to capture all logs in one single file.
- [x] Right now the extra runs created will blow up the UI, and also, "workspace/latest" won't have the outputs for the main multicore run. Potentially use this PR: https://github.com/allenai/tango/pull/212 to refactor the code so that all steps belong to a single run despite running in different processes.
- [x] Logs from child processes of steps that contain multiprocessing themselves _are_ captured, but they don't have the step_name prefix. Look into why that happens.
- [x] Make `MulticoreExecutor` the default.

To be done in a separate PR: https://github.com/allenai/tango/issues/223

## Before submitting

## After submitting


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `tango.settings` module and `tango settings` group of commands.
 - A format for storing sequences as `SqliteSparseSequence`
 - A way to massage kwargs before they determine the unique ID of a `Step`
+- Added CLI option to run a single step in a config using `--step-name` or `-s`.
```
epwalsh, line 33: Should be under "Unreleased" ⬆️

tango/__init__.py
```diff
@@ -15,7 +15,7 @@
     "Workspace",
 ]
 
-from tango.executor import Executor
+from tango.executors import Executor
```
epwalsh, line 18: For consistency with `Step`, `Workspace`, `StepCache`, etc., I think we should keep the base class in `tango/executor.py`, and then put subclasses in `tango/executors/*`.

tango/__main__.py
```diff
@@ -612,9 +641,18 @@ def _run(
...
     params.assert_empty("'tango run'")
-    run = workspace.register_run(step for step in step_graph.values())
+
+    if step_name is not None:
+        assert step_name in step_graph, (
+            f"You want to run a step called '{step_name}', but it cannot be found in the experiment config. "
+            f"The config contains: {list(step_graph.keys())}."
+        )
+        sub_graph = step_graph.get_sub_graph(step_name)
+        run = workspace.register_run(step for step in sub_graph.values())
```
epwalsh, line 653: Does this mean that a new run will be created for each step that `MulticoreExecutor` runs?

epwalsh (commented):
Looks like a great start! Just a couple minor comments and questions so far. I'll dig more into the details soon.

tango/__main__.py
epwalsh: If so, I think we want to avoid that. That would definitely create issues with the `WandbWorkspace`.

AkshitaB: Yes, this is on my TODO to fix. Ideally, we want a single registered run, and the other processes "editing" that run.

tango/step_graph.py
```diff
@@ -151,28 +185,77 @@ def __iter__(self) -> Iterator[str]:
...
-                steps_run.add(step)
-                result.append(step)
-            else:
-                steps_not_run.append(step)
+        for step_name in cls._get_ordered_steps(dependencies):
+            step_dict[step_name].name = step_name
+            result.append(step_dict[step_name])
         return result
 
+    def find_uncacheable_leaf_steps(self) -> Set[Step]:
```
dirkgr, line 206: Let's be consistent about this kind of naming. It's not `Step.get_workspace()`, it's just `Step.workspace()`. So it should just be `uncacheable_leaf_steps()` here, like it already is for `ordered_steps()` above. Same for `sub_graph()`.

tango/step_graph.py
```diff
@@ -151,28 +185,77 @@ def __iter__(self) -> Iterator[str]:
...
+                step_dict[step_name] = {
+                    key: _to_config(val) for key, val in step._to_params()["kwargs"].items()
+                }
+                step_dict[step_name]["type"] = step.__module__ + "." + step.__class__.__name__
+
+                # We only add cache_results and format to the config if the values are different from default.
+                if step.cache_results != step.CACHEABLE:
+                    step_dict[step_name]["cache_results"] = step.cache_results
+                if step.format != step.FORMAT:
+                    step_dict[step_name]["step_format"] = _to_config(step.format._to_params())
```
dirkgr, line 254: Nice catch. Did you run into a problem caused by this, or did you just know? I would have forgotten.

dirkgr: But wait, shouldn't this happen in `Step._to_params()`?

dirkgr (changes requested):

tango/executors/executor.py
```diff
@@ -24,30 +24,30 @@ def __init__(
...
+            self.execute_step(step, step in uncacheable_leaf_steps)
 
-        for step in ordered_steps:
-            if step.cache_results:
-                step.ensure_result(self.workspace)
-            elif step in uncacheable_leaf_steps:
-                step.result(self.workspace)
+    @property
+    def failed_steps(self) -> Set[str]:
+        return set()
```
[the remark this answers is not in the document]
AkshitaB: Each call to `execute_step_graph` resets the failed steps in `MulticoreExecutor`. And we don't have a way to run more steps in `SimpleExecutor` if a step fails, so the above isn't used in that case.

tango/executors/executor.py
```diff
@@ -24,30 +24,30 @@ def __init__(
...
 
         # Import included packages to find registered components.
         if self.include_package is not None:
             for package_name in self.include_package:
                 import_extra_module(package_name)
 
+        if step.cache_results:
+            step.ensure_result(self.workspace)
+        elif is_uncacheable_leaf_step:
+            step.result(self.workspace)
```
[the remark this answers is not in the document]
AkshitaB: Yeah, for now, I'm just sticking to how we have been doing it.

tango/__main__.py
```diff
@@ -640,13 +677,16 @@ def _run(
...
+        executor: Executor
         # Initialize Executor and execute the step graph.
-        executor = Executor(workspace=workspace, include_package=include_package)
-        executor.execute_step_graph(step_graph)
+        if multicore:
+            executor = MulticoreExecutor(
+                workspace=workspace, include_package=include_package, parallelism=parallelism
+            )
+        else:
+            executor = Executor(workspace=workspace, include_package=include_package)
```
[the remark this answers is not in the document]
AkshitaB: Yes.

tango/step_graph.py
AkshitaB: It occurred to me to add this check because I was using `cache_results=False` to test something else earlier. Yes, I think this can be moved to `Step._to_params()` -- I think I wanted to isolate file changes because this is the only place where this was being used.

tango/executors/multicore_executor.py
```diff
@@ -0,0 +1,250 @@
...
+
+        assert not self._running and not self._queued_steps
+
+    @property
+    def failed_steps(self) -> Set[str]:
+        return self._failed
+
+    def _has_incomplete_steps(
+        self, step_graph: StepGraph, step_states: Dict[str, StepState]
+    ) -> bool:
```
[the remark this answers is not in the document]
AkshitaB: I make a distinction because `_has_incomplete_steps` needs to wait for steps that are currently running as well, whereas, `_get_steps_to_run` will only give me the steps that can be run immediately (dependencies are satisfied, etc.).

tango/format.py
```diff
@@ -115,13 +122,14 @@ class DillFormat(Format[T], Generic[T]):
 
     def __init__(self, compress: Optional[str] = None):
         try:
-            self.open = _OPEN_FUNCTIONS[compress]
+            self.compress = compress
         except KeyError:
             raise ConfigurationError(f"The {compress} compression format does not exist.")
```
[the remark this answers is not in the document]
AkshitaB: Hmm, I like that idea. Every `FromParams` object would know how exactly it was constructed.

tango/executors/multicore_executor.py
```diff
@@ -0,0 +1,250 @@
...
+    def _get_steps_to_run(
+        self, step_graph: StepGraph, step_states: Dict[str, StepState]
+    ) -> Set[str]:
+        """
+        Returns the steps that can be queued to run immediately.
+        Criteria:
+            1) All dependencies are available.
+            2) Step is not already running or queued.
+            3) Step has not run in the past and failed.
+            4) Step's state is INCOMPLETE
```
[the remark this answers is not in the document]
AkshitaB: Yes, I check it in code; will update the comment.

tango/executors/multicore_executor.py
```diff
@@ -0,0 +1,250 @@
...
+
+        (We only run uncacheable steps if they are needed for another step downstream,
+        as part of the downstream step).
+        """
+
+        def _are_dependencies_available(step: Step) -> bool:
+            for dependency in step.dependencies:
+                if step_states[dependency.name] not in [StepState.COMPLETED, StepState.UNCACHEABLE]:
```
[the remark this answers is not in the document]
AkshitaB: Not sure what you mean. This is checking if the step's dependency is completed or uncacheable (in which case, we can directly run the step). Whether or not the step itself is cacheable doesn't matter. We don't want to run a step if its dependencies cannot be met.

tango/__main__.py
```diff
@@ -173,11 +188,17 @@ def main(
     if file_friendly_logging:
         settings.file_friendly_logging = file_friendly_logging
 
-    initialize_logging(
-        log_level=settings.log_level,
-        file_friendly_logging=settings.file_friendly_logging,
-        enable_click_logs=True,
-    )
+    if called_by_executor:
+        # We only set this flag instead of calling `initialize_prefix_logging` here
```
dirkgr, line 192: `initialize_prefix_logging`?

tango/__main__.py
```diff
@@ -242,13 +277,16 @@ def run(
...
+    parallelism: int = 1,
+    step_name: Optional[str] = None,
     name: Optional[str] = None,
 ):
     """
     Run a tango experiment.
 
     EXPERIMENT is the path to experiment's JSON/Jsonnet/YAML configuration file.
     """
+    global _CALLED_BY_EXECUTOR
```
dirkgr, line 289: If you're just reading the value, you don't need to declare it `global`.

tango/__main__.py
```diff
@@ -656,9 +732,48 @@ def _run(
...
+            log_step_summary(step)
+            click_logger.info(
+                click.style("Finished run for step ", fg="green")
+                + click.style(step_name, fg="green", bold=True)
+                + click.style(" (", fg="green", bold=True)
+                + click.style(run.name, fg="green", bold=True)
+                + click.style(")", fg="green", bold=True)
+            )
+        else:
+            failed_steps = executor.execute_step_graph(step_graph, run_name=run.name)
```
dirkgr, line 748: It seems weird to me that `execute_step_graph()` returns failed steps. That's what you need here, sure, but `execute_step_graph()` is an incredibly general function, so it's weird to tailor the return type to one use case. At least, it should return a tuple `(succeeded_steps, failed_steps, neither_steps)`.

tango/__main__.py
```diff
@@ -656,9 +732,48 @@ def _run(
...
+            )
+
+    if called_by_executor:
+        from tango.common.aliases import EnvVarNames
+
+        # We set this environment variable so that any steps that contain multiprocessing
+        # and call `initialize_worker_logging` also log the messages with the `step_name` prefix.
+        os.environ[EnvVarNames.LOGGING_PREFIX.value] = f"step {step_name}"
+        initialize_prefix_logging(f"step {step_name}", main_process=False)
+        log_and_execute_run()
```
dirkgr, line 772: Can we use this method of initializing logging all the time, instead of as a one-off for `MulticoreExecutor`?

I don't have the overview over all the logging scenarios, so take my opinions on the matter with a grain of salt. All I see is some code smells in the `called_by_executor` flag, the global state, and the one-off code path. I fear we're fairly committed to doing opaque env var magic at this point. We already do that in other cases. But do we have the one-off code paths? Can we just use the prefix stuff all the time?

tango/executor.py
```diff
@@ -24,30 +27,26 @@ def __init__(
...
+    def execute_step(self, step: "Step", is_uncacheable_leaf_step: bool = False) -> None:
         # Import included packages to find registered components.
         if self.include_package is not None:
             for package_name in self.include_package:
                 import_extra_module(package_name)
 
+        if step.cache_results:
+            step.ensure_result(self.workspace)
+        elif is_uncacheable_leaf_step:
+            step.result(self.workspace)
```
dirkgr, line 39: Why do we have to check this? Whether it's leaf or not doesn't matter here, does it? When I call a method called `execute_step()`, I expect that either the step gets executed, or I get an exception. Under no circumstances would I expect the method to return after doing nothing.

tango/executors/executor.py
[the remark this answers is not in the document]
dirkgr: 👍🏻

tango/format.py
```diff
@@ -63,6 +63,13 @@ def read(self, dir: PathOrStr) -> T:
         """Reads an artifact from the directory at ``dir`` and returns it."""
         raise NotImplementedError()
 
+    def _to_params(self) -> Dict[str, Any]:
+        params_dict = super()._to_params()
+        for key in ["logger", "__orig_class__"]:
+            params_dict.pop(key, None)  # Removing unnecessary keys.
+        params_dict["type"] = self.__module__ + "." + self.__class__.__name__
```
dirkgr, line 70: I think this has to be `__qualname__` instead of `__name__`.

dirkgr (changes requested):
Didn't you want to put some `FromParams` refactoring in?

tango/executor.py
AkshitaB: This is how it is currently run, and I have not changed the logic. Perhaps you can shed some light on [this](https://github.com/allenai/tango/commit/c46e0b2200012aad12e7ed232a3b076d979abeb9)? Perhaps it was needed then, and is now just a relic?

tango/__main__.py
AkshitaB: For the single process executor, there's no point in calling `initialize_prefix_logging`, as the logging has already been initialized once, for the main process (and it is only being called for `step_name`; which we won't have in the general case). I don't particularly like the use of global env variables myself, but we already do it for the rest of the multiprocessing-friendly logging. 

The reason why I added `called_by_executor` was because I wanted to accommodate the case for people running a single step themselves without it necessarily being part of the `MulticoreExecutor` run. 

There were some issues related to sockets if I tried to just run `initialize_prefix_logging` all the time. I can spend some more time on it.

AkshitaB:
> Didn't you want to put some `FromParams` refactoring in?

@dirkgr We decided to do that in a separate PR. I've made an issue for it here: https://github.com/allenai/tango/issues/223

tango/executors/multicore_executor.py
```diff
@@ -0,0 +1,237 @@
...
+        self.include_package = include_package
+        self.parallelism = parallelism
+
+        # Perhaps there's a better way to do this without these being passed as args.
+        self._num_tries_to_sync_states = num_tries_to_sync_states
+        self._wait_seconds_to_sync_states = wait_seconds_to_sync_states
+
+    def execute_step_graph(self, step_graph: StepGraph, run_name: Optional[str] = None) -> Set[str]:
+        """
+        Execute a :class:`tango.step_graph.StepGraph`.
```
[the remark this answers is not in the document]
AkshitaB: I have implemented it according to policy 2. If something fails, we don't run anything that requires that step, but we _do_ run unrelated steps. This makes sense to me because say, I kick off some long processing steps overnight, and the first one fails -- it would be nice to have the remaining ones still finished by the morning. I will add this to the docstring, but let me know if you are strongly in favor of policy 1.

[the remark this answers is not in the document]
dirkgr: No, makes sense to me. Policy 2 it is!

tango/executor.py
dirkgr: Before your refactor, that code's responsibility was to execute a whole step graph. When you're doing that, it makes sense that you execute all cacheable steps, and also all uncacheable leaf steps. Interior uncacheable steps will be run automatically, but for uncacheable leaf steps we have to make sure they run manually, because they won't be run as part of some other step's dependencies.

Now, the name of the method promises that it will run a single step. In that context, it makes no sense to make that distinction.

tango/executors/multicore_executor.py
```diff
@@ -0,0 +1,237 @@
...
+        _failed: Set[str] = set()
+        _queued_steps: List[str] = []
+
+        def _sync_step_states() -> Dict[str, StepState]:
+            """
+            Update the StepState info.
+            Although, this is not really elegant. The issue is as follows: The main multicore executor process
+            queues a step, and starts step execution in a different process. If we try to read the StepState
+            before that process has had time to update the StepState, the Workspace will throw the out of sync
+            error (IOError: process should be running but it's considered incomplete...).
```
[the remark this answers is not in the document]
AkshitaB: Yes, I can solve it this way, but there's an edge case that I ran into: 

[This](https://github.com/allenai/tango/blob/main/test_fixtures/package/steps.py#L40) step is supposed to be non-deterministic, but according to Tango, it's not. It's for our test case, so we can change it, but users can write what they want. 

And we have [this](https://github.com/allenai/tango/blob/main/test_fixtures/experiment/random.jsonnet) config with: 
```
"steps": {
        "rand_string1": {"type": "random_string"},
        "rand_string2": {"type": "random_string"},
        ...
```

When calling `workspace.step_info(step)` to sync the states for each step, both these steps will be considered the same (as according to Tango, they are deterministic), so if `rand_string1` has been kicked off, and we don't check it, we will run into an error when checking `rand_string2`'s step state. With the current implementation, it will gracefully find that both steps have finished, but if I only check `_running`, it will fail. 

We may not want to support this edge case since, technically, there should be no need to run a deterministic step twice. 

The other option (which makes more sense, I guess) is to explicitly check if any steps are the same (but with different names), in which case, we warn the user, and remove the duplicate step from the run's config, and update the dependencies. Or just plainly fail without running anything if we find a duplicate step.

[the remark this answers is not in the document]
AkshitaB: In some ways, this mimics the issue that will occur with concurrent runs, so I think it still makes sense to keep the current implementation.

tango/executor.py
```diff
@@ -27,26 +45,39 @@ def __init__(
...
         """
 
-        ordered_steps = sorted(step_graph.values(), key=lambda step: step.name)
-        uncacheable_leaf_steps = step_graph.uncacheable_leaf_steps()
+        successful: Set[str] = set()
+        failed: Set[str] = set()
+        not_run: Set[str] = set()
+        error_tracebacks: List[str] = []
+        for step in step_graph.values():
+            if any([dep.name in failed for dep in step.recursive_dependencies]):
```
dirkgr, line 70: You don't need the extra `[]` there. `any()` can take a generator, and it'll save one list allocation.

tango/executor.py
```diff
@@ -27,26 +45,39 @@ def __init__(
...
         # Import included packages to find registered components.
         if self.include_package is not None:
             for package_name in self.include_package:
                 import_extra_module(package_name)
 
-        if step.cache_results:
-            step.ensure_result(self.workspace)
-        elif is_uncacheable_leaf_step:
-            step.result(self.workspace)
+        step.result(self.workspace)
```
dirkgr, line 54: This should still call `ensure_result()` for cacheable steps. That's an optimization that can potentially save a lot of disk IO.

tango/executor.py
```diff
@@ -27,26 +45,39 @@ def __init__(
...
+        for step in step_graph.values():
+            if any([dep.name in failed for dep in step.recursive_dependencies]):
+                not_run.add(step.name)
+            else:
+                try:
+                    self.execute_step(step)
+                    successful.add(step.name)
+                except Exception:
+                    failed.add(step.name)
+                    error_tracebacks.append(traceback.format_exc())
```
dirkgr, line 78: Uncacheable interior steps should not be executed explicitly like this. Otherwise they will run one extra time for no reason.

dirkgr: Does the multicore executor do this correctly?

tests/executor_test.py
```diff
@@ -43,6 +44,39 @@ def test_executor(self):
...
+                    "type": "sleep-print-maybe-fail",
+                    "string": {"type": "ref", "ref": "failing_step"},
+                    "seconds": 0,
+                    "fail": False,
+                },
+            }
+        )
+        executor = Executor(workspace)
+        assert len(executor.workspace.step_cache) == 0
+        output = executor.execute_step_graph(step_graph)
```
dirkgr, line 78: When looking at it like this, this seems a little dangerous. It's easy to accidentally write code the swallows errors. It should be very clear in the docs that `execute_step_graph()` will not propagate exceptions.

dirkgr (changes requested):

tango/common/aliases.py
```diff
@@ -12,6 +12,7 @@ class EnvVarNames(Enum):
     CLICK_LOGGER_ENABLED_ENV_VAR = "TANGO_CLICK_LOGGER_ENABLED"
     LOGGING_HOST_ENV_VAR = "TANGO_LOGGING_HOST"
     LOGGING_PORT_ENV_VAR = "TANGO_LOGGING_PORT"
+    LOGGING_PREFIX = "LOGGING_PREFIX"
```
epwalsh, line 15: For consistency with the other env var names and to make this more "tango-specific", I think we should prefix by `TANGO_`:
```suggestion
    LOGGING_PREFIX = "TANGO_LOGGING_PREFIX"
```

dirkgr (approved):
I think this is it?

PR #205, opened 2022-02-22
Robustness fixes

dirkgr:
- A format for storing sequences as `SqliteSparseSequence`.
- `MappedSequence` now works with slicing.
- `MappedSequence` now works with Huggingface `Dataset`.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -17,7 +17,6 @@ isort==5.10.1
 # Running tests
 pytest
 pytest-sphinx
-datasets  # We have a test that needs this.
 
 # Flaky tests
 flaky
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -37,7 +37,7 @@ wheel
 Sphinx==4.4.0
 
 # Sphinx theme: https://sphinx-themes.org/sample-sites/furo/
-furo==2022.1.2
+furo==2022.2.14.1
 
 # Lets Sphinx parse markdown files in addition to rst.
 myst-parser==0.17.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -17,6 +17,7 @@ isort==5.10.1
 # Running tests
 pytest
 pytest-sphinx
+datasets  # We have a test that needs this.
 
 # Flaky tests
 flaky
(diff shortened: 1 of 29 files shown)


dev-requirements.txt
```diff
@@ -17,6 +17,7 @@ isort==5.10.1
 # Running tests
 pytest
 pytest-sphinx
+datasets  # We have a test that needs this.
```
epwalsh, line 20: To keep our dev-dependencies light, that test could be put into `tests/integrations/datasets` instead.

dirkgr: Hmm. It's a test for `MappedSequence`. You have to special case when the sequence is a `Dataset`, because `Dataset`s look like sequences, but they are not. It looks and quacks like a duck, but it's actually a platypus.

I'll do it anyways.

tango/format.py
```diff
@@ -401,6 +402,29 @@ def __next__(self) -> str:
             raise StopIteration()
 
 
+@Format.register("sqlite_sequence")
+class SqliteSequenceFormat(Format[Sequence[T]]):
+    VERSION = 3
+
+    FILENAME = "data.sqlite"
```
epwalsh, line 409: I feel like all `Format`s should make their file name or directory name transparent like this. Just a thought.

epwalsh (approved):
LGTM

tango/format.py
dirkgr: I don't think that's always possible. Some formats, like the sqlite dataset dict format, need the directory structure. The sqlite dataset format stores the split names in filenames!

epwalsh: Yes, same with `DatasetsFormat`, but in that case they could show the directory name. So we could call this `PATH` instead of `FILENAME`.

dirkgr: There is no directory name though, from the perspective of the `Format` class. The format class is given a directory, and it should write into that directory. Sometimes it writes only one file, but sometimes it writes many.

epwalsh: My point is that it can write it into a subdirectory of the given dir. This is how `DatasetsFormat` works.

dirkgr: I don't know what all `Dataset.save_to_disk()` writes in there, but it should not create extra subdirectories if it doesn't have to. I think people will want to go into these directories and mess with the files themselves. This is, after all, the way we write out results. So those directories should be somewhat human readable.

PR #209, opened 2022-02-24
Allow setting environment variables in tango settings, fix bug with TANGO_LOG_LEVEL env var

epwalsh:
Adds a field `environment: Dict[str, str]` to `TangoGlobalSettings`, which is a mapping of env variables that will be added to the environment each time `tango` is run. Also adds a subcommand `tango settings set env` to add or update these variables.

I've found this helpful when working on the W&B Workspace, because there are certain W&B environment variables that I want to set when using tango, but not otherwise.

Any environment variables are allowed to be set this way EXCEPT for those that already have a corresponding field in `TangoGlobalSettings`, such as `TANGO_LOG_LEVEL` which has the corresponding settings field `log_level`. There are two reasons we should disallow setting those specific env vars that way:
1. It would lead to confusing behavior because it's not obvious which should take precedence, the env var or the settings field.
2. The global variables that take values from those environment variables (e.g. `tango.common.logging.LOG_LEVEL`) will have already been imported by the time we get to the point where we're setting those environment variables from `TangoGlobalSettings.environment`, so they would take no effect unless we refactor how we initialize those global variables.

This also fixes a bug where the `TANGO_LOG_LEVEL` env var was completely ignored.

### DIFF
--- a/tango/common/aliases.py
+++ b/tango/common/aliases.py
@@ -1,12 +1,18 @@
+from enum import Enum, unique
 from os import PathLike
-from typing import Union
+from typing import Set, Union
 
 PathOrStr = Union[str, PathLike]
 
 
-class EnvVarNames:
+@unique
+class EnvVarNames(Enum):
     FILE_FRIENDLY_LOGGING_ENV_VAR = "FILE_FRIENDLY_LOGGING"
-    LOG_LEVEL_ENV_VAR: str = "TANGO_LOG_LEVEL"
-    CLICK_LOGGER_ENABLED_ENV_VAR: str = "TANGO_CLICK_LOGGER_ENABLED"
-    LOGGING_HOST_ENV_VAR: str = "TANGO_LOGGING_HOST"
-    LOGGING_PORT_ENV_VAR: str = "TANGO_LOGGING_PORT"
+    LOG_LEVEL_ENV_VAR = "TANGO_LOG_LEVEL"
+    CLICK_LOGGER_ENABLED_ENV_VAR = "TANGO_CLICK_LOGGER_ENABLED"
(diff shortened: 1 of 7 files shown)


tango/common/aliases.py
```diff
@@ -2,3 +2,11 @@
...
 
 PathOrStr = Union[str, PathLike]
+
+
+class EnvVarNames:
+    FILE_FRIENDLY_LOGGING_ENV_VAR = "FILE_FRIENDLY_LOGGING"
+    LOG_LEVEL_ENV_VAR: str = "TANGO_LOG_LEVEL"
+    CLICK_LOGGER_ENABLED_ENV_VAR: str = "TANGO_CLICK_LOGGER_ENABLED"
+    LOGGING_HOST_ENV_VAR: str = "TANGO_LOGGING_HOST"
+    LOGGING_PORT_ENV_VAR: str = "TANGO_LOGGING_PORT"
```
dirkgr, line 12: I'm not super keen on this structure. Seems unnecessary complex, and then you have to use `__dict__` to take advantage of it. What's wrong with just a `Set[str]`?

I know in CS 101 they teach that you should never use raw literals, but in cases like this it buys you nothing. What's the advantage of always typing `MyClass.MY_LONG_NAME_VAR` instead of `"MY_LONG_NAME"`? I guess you get some resistance against typos. That's a pretty weak gain for abusing the class system like this.

Anyways, not a ship blocker or anything, but I don't like it.

dirkgr (approved):

tango/common/aliases.py
epwalsh: Yea, you're right. That's not great. I switched to using an Enum instead: fe943e5

Issue #181, opened 2022-02-09
Default sweeper class to read hyperparameters


SabhyaC26:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #142, opened 2022-01-19
Allow specifying specific "target" steps in the graph to run in the `Executor` and from the `tango run` command


epwalsh:
Let's say we have a config like this:

```jsonnet
// config.jsonnet
{
  steps: {
    dataset: {},
    trained_model: {}, // depends on "dataset"
    eval_results: {},  // depends on "trained_model"
  }
}
```

But sometimes we only want the `trained_model` and don't care about `eval_results`. With this feature, you could do

```
tango run config.jsonnet --step trained_model
```

This would result in only running `dataset` (because it's an input to `trained_model`) and `trained_model`, but `eval_results` is not run.

PR #214, opened 2022-03-01
Sweeper PR

### LINKED ISSUE #143: Hyperparameter sweeps
It would be great if Tango provided a simple yet general mechanism for doing hyperparameter searches. Here's an outline of how that could look 👇

We provide a new subcommand: `tango sweep`. This command takes
- a "sweep" config,
- a regular experiment config,
- and a target step name.

For example:

```
tango sweep sweep-config.jsonnet target-config.jsonnet step-name
```

`step-name` should correspond to the main step of interest in `target-config.jsonnet` that provides the results we are trying to optimize for (might require #142). For example, this could be a validation/eval step that spits out some metrics of your model on a dataset.

The `sweep-config.jsonnet` would define which hyperparameters to search and how to search over them. By "hyperparemeters" I really just mean any fields in `target-config.jsonnet`. There are many ways we could do the search, and this is an active area of research. So I think it would be ideal if we were able to integrate with existing hyperparameter sweep frameworks / platforms, like [W&B](https://docs.wandb.ai/guides/sweeps), [Optuna](https://optuna.org/), etc. These integrations should be optional, however, and I think we should provide a simple default search method, which could just be grid search.

Under the hood `tango sweep` could use the `tango run` subcommand with the `--overrides` parameter to select hyperparameter values. We should also be able to run the search in parallel.

SabhyaC26:
Fixes #143 #181

- new sweeper class for performing hyperparameter sweeps in tango

## Before submitting

## After submitting


tango/sweeps.py
```diff
@@ -0,0 +1,67 @@
...
+sweeps_config_path = (
+    "/Users/sabhyac/Desktop/sabhya/tango/test_fixtures/sweeps/basic_test/sweeps-config.jsonnet"
+)
+components = (
+    "/Users/sabhyac/Desktop/sabhya/tango/test_fixtures/sweeps/basic_test/basic_arithmetic.py"
+)
+
+
+class Sweeper(Registrable):
+    def __init__(self, main_config_path: str, sweeps_config_path: str, components: str):
```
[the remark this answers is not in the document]
epwalsh: You may want to start by implementing https://github.com/allenai/tango/issues/142 in a separate PR

tango/sweeps_test.ipynb
```diff
@@ -0,0 +1,393 @@
+{
```
epwalsh, line 1: Can you remove this from the PR? You can tell git to ignore this file by adding it to your global git ignore

tango/sweeps.py
```diff
@@ -0,0 +1,67 @@
...
+
+main_config_path = (
+    "/Users/sabhyac/Desktop/sabhya/tango/test_fixtures/sweeps/basic_test/config.jsonnet"
+)
+sweeps_config_path = (
+    "/Users/sabhyac/Desktop/sabhya/tango/test_fixtures/sweeps/basic_test/sweeps-config.jsonnet"
+)
+components = (
+    "/Users/sabhyac/Desktop/sabhya/tango/test_fixtures/sweeps/basic_test/basic_arithmetic.py"
+)
```
epwalsh, line 16:
```suggestion
```

epwalsh (changes requested):

tango/sweeps.py
```diff
@@ -0,0 +1,68 @@
...
+
+from tango.common import Params, Registrable
+
+
+class Sweeper(Registrable):
+    def __init__(
+        self,
+        main_config_path: str,
+        sweeps_config_path: str,
+        components: str,
```
epwalsh, line 15: The `Sweeper` class needs to be aware of the `Workspace` being used in order to get the results of each target step, and to pass the workspace URL when it runs each experiment (with the `-w` option).

tango/sweeps.py
```diff
@@ -0,0 +1,68 @@
...
+    ):
+        super(Registrable, self).__init__()
+        self.main_config_path = main_config_path
+        self.sweep_config = load_config(sweeps_config_path)
+        self.main_config_path = main_config_path
+        self.components = components
+
+    # returns all the combinations of hyperparameters in the form of a list of lists
+    def get_combinations(self) -> list:
```
epwalsh, line 24: For compat with older Python versions:
```suggestion
    def get_combinations(self) -> List:
```

epwalsh, line 24: It would probably be better if this method was a generator function so that implementations could dynamically pick the next combination based on the history so far.

tango/sweeps.py
```diff
@@ -0,0 +1,68 @@
...
+                    self.main_config_path,
+                    "--include-package",
+                    self.components,
+                    "--overrides",
+                    json.dumps(overrides),
+                ]
+            )
+
+    # function to override all the hyperparameters in the current experiment_config
+    def override_hyperparameters(self, experiment_tuple: tuple) -> dict:
```
epwalsh, line 53: Docstrings should be written like this:
```suggestion
    def override_hyperparameters(self, experiment_tuple: tuple) -> dict:
        """
        Generates a overrides dictionary for the current experiment config.
        """
```

tango/sweeps.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Sweeper(Registrable):
+    def __init__(
+        self,
+        main_config_path: str,
+        sweeps_config_path: str,
+        components: str,
+    ):
+        super(Registrable, self).__init__()
```
epwalsh, line 17: In recent versions of Python you can just do this:
```suggestion
        super().__init__()
```

epwalsh (changes requested):

Issue #219, opened 2022-03-04
HF ecosystem docs no longer create object.inv file needed by Intersphinx


epwalsh:
Hopefully https://github.com/huggingface/doc-builder/issues/81 gets resolved, otherwise we'll have to either (A) find a new reliable way of linking to HF API docs, or (B) don't link to HF API docs.

PR #220, opened 2022-03-04
Temporarily remove intersphinx links to HF docs

epwalsh:
Temporarily fix for #219 so our doc builds won't fail.


epwalsh:
Thanks @schmmd!

PR #231, opened 2022-03-15
Soft prompts

dirkgr:
In code:

```Python
import transformers
t = transformers.AutoModel.from_pretrained("gpt2")
twp = make_prefix_transformer(t, prefix_length=3)
```

In config files:
```Jsonnet
{
    model: {
        type: "transformers::with_soft_prompt",
        prompt_length: 3,
        model: {
            type: "transformers::AutoModelForCausalLM::from_pretrained",
            pretrained_model_name_or_path: "gpt2"
        },
    }
}
```

Missing:
 - [x] Tests
 - [x] Docs
 - [x] Try it with T5
 - [x] A proper end-to-end training config that uses this
 - [x] Add an easy way to make only the prefix trainable, and leave the rest of the weights alone


tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+                "Unexpected result type from the transformer in prefix_transformer: `%s`",
+                result.__class__,
+            )
+            return result
+
+    model.forward = new_forward
+    return model
+
+
+Model.register("transformers::with_prefix")(make_prefix_transformer)
```
ZhaofengWu, line 111: As this will probably be difficult to change later, it's worth thinking about the terminology. Prefix tuning? Prompt tuning? Something else?

tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+        # following Lester et al. 2021 in initializing using the top 5000 random vocabs
+        r = random.Random(random_seed)
+        indices = r.sample(range(5000), prefix_length)
+        with torch.no_grad():
+            self.prefix_embedding.weight.copy_(self.original_embedding.weight[indices])
+
+    def forward(self, input: torch.Tensor):
+        embedded_prefix = self.prefix_embedding(input[:, : self.prefix_length])
+        embedded_rest = self.original_embedding(input[:, self.prefix_length :])
+        return torch.cat([embedded_prefix, embedded_rest], dim=1)
```
ZhaofengWu, line 35: This makes the assumption that the prefix comes before the text. This is probably fine and should work in most cases, but just wanted to note that my version doesn't make this assumption.

tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+    assert isinstance(model, PreTrainedModel)
+
+    model.set_input_embeddings(WithPrefixEmbedding(model.get_input_embeddings(), prefix_length))
+
+    def patch_tensor(kwargs: Dict[str, torch.Tensor], key: str, value: Any = 0) -> None:
+        try:
+            t = kwargs[key]
+        except KeyError:
+            return
+        zero_prefix = torch.full(
```
ZhaofengWu, line 48: `full_like`?

tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+        return torch.cat([embedded_prefix, embedded_rest], dim=1)
+
+
+def make_prefix_transformer(model: Model, prefix_length: int) -> Model:
+    assert isinstance(model, PreTrainedModel)
+
+    model.set_input_embeddings(WithPrefixEmbedding(model.get_input_embeddings(), prefix_length))
+
+    def patch_tensor(kwargs: Dict[str, torch.Tensor], key: str, value: Any = 0) -> None:
+        try:
```
ZhaofengWu, line 44: Why can't you test with `key in kwargs`?

tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+            t = kwargs[key]
+        except KeyError:
+            return
+        kwargs[key] = torch.cat(
+            [
+                torch.arange(0, prefix_length, dtype=t.dtype)
+                .unsqueeze(0)
+                .expand(t.size(0), prefix_length),
+                t + prefix_length,
```
ZhaofengWu, line 63: Do you need the `+ prefix_length`? The original token indices are not changed, right?

tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+            ],
+            dim=1,
+        )
+
+    # Because PyTorch hooks don't support kwargs, we monkey patch the forward method 🙈
+    old_forward = model.forward
+
+    def new_forward(*args, **kwargs):
```
ZhaofengWu, line 71: What made me turn away from monkeypatching in my code is when I noticed that it doesn't need/have a `self`, so there might be some fundamental differences between the old vs. new forward. If I were two years younger I probably would have voted for monkeypatching but the older me is less adventurous and worry more about safety. Go ahead if you're confident that this is safe, but at least I would suggest some sort of assertion to check the forward has not been monkeypatched (because if it had, the logic would be incorrect).

tango/integrations/transformers/prefix_transformer.py
```diff
@@ -0,0 +1,111 @@
...
+
+        patch_tensor_with_indices(kwargs, "input_ids")
+        patch_tensor(kwargs, "labels")
+        patch_tensor(kwargs, "attention_mask", 1)
+        patch_tensor(kwargs, "token_type_ids")
+        patch_tensor_with_indices(kwargs, "position_ids")
+
+        result = old_forward(*args, **kwargs)
+
+        if isinstance(result, CausalLMOutputWithCrossAttentions):
```
ZhaofengWu, line 84: Some comment for what this is doing?

tango/integrations/transformers/prefix_transformer.py
dirkgr: This is true, but this is not the place where that assumption is made. The assumption is made when we create the `position_id` input. It could be easily changed there.

tango/integrations/transformers/prefix_transformer.py
dirkgr: I can, but then it's two lookups into `kwargs` when one is sufficient.

tango/integrations/transformers/prefix_transformer.py
dirkgr: Shape is different

tango/integrations/transformers/prefix_transformer.py
dirkgr: You are right. This is a bug for `input_ids`. I wrote the code for `position_ids` and then got sloppy.

tango/integrations/transformers/prefix_transformer.py
dirkgr: I am a little uneasy about it, but I think it beats the alternatives. At the very least I want to see where it goes and where it falls down, if it does. Also, apparently there is movement on the PyTorch side to allow `kwargs` in hooks. When that comes true, we can do this properly.

As for your specific concern, this will work fine even if `forward()` has already been monkey patched before.

tango/integrations/transformers/prefix_transformer.py
dirkgr: Yeah, I have to go through and write docs and whatnot.

tango/integrations/transformers/prefix_transformer.py
dirkgr: Maybe we'll call it `with_soft_prompt`?

tango/integrations/transformers/prefix_transformer.py
ZhaofengWu: Will it? Wouldn't be the patching happen multiple times at each recursion level?

tango/integrations/transformers/prefix_transformer.py
ZhaofengWu: `new_full` then

ZhaofengWu: Though I don't particularly like the un-readability of `new_full`

tango/integrations/transformers/prefix_transformer.py
dirkgr: If the thing you pass into this function was monkey patched before, then `old_forward` ends up being the first level of monkey patching, and it will get called when we go one level down.

`old_forward` becomes part of the closure of `new_forward`. That's how the chain of forward methods is maintained.

tango/integrations/transformers/prefix_transformer.py
dirkgr: Ah, `new_full` is indeed better. Readability concerns aside, I like that when I use a method like that, it'll just work when the torch devs think of a new piece of metadata that could be attached to a tensor.

tango/integrations/transformers/prefix_transformer.py
ZhaofengWu: Right. So it would be like this, no?
```
forward:  # monkey patch lvl 1
  patch_tensor
  forward:  # monkey patch lvl 2  
    patch_tensor
    forward  # original
```

tango/integrations/transformers/prefix_transformer.py
dirkgr: 0cc761f

tango/integrations/transformers/prefix_transformer.py
dirkgr: Yes

ZhaofengWu: So the tensors will be patched twice

dirkgr: The inputs and outputs should be patched twice. That is correct.

What won't work is that you can't call `set_input_embeddings()` twice with the way I have it here, because `_WithPromptEmbedding` reaches into the original embedding's internals.

ZhaofengWu: Oh, what I was worried about is the method being unintentionally called twice. I can't think of a case where it is intentionally called twice.

dirkgr: c8d1b86 should make it possible to stack two prompt-enabled transformers on top of each other.

I think it's important that we ensure this pattern works for other forms as well. What if we implement adapters the same way, and we want to run both at the same time? The whole point of trying for this "looks like a normal huggingface transformer" approach is that it should be easy to combine with other components that do the same thing.

dirkgr:
Oh no, I found a big problem with this. It doesn't work with `past_key_values`. Fix incoming.

dirkgr:
This does not work for T5 at all 😭. I'm no longer sure this approach of patching the model will work. The huggingface generation code makes calls into the middle of their model, instead of always going through the `forward()` method. So patching `forward()` doesn't work. And patching `forward()` of an internal method breaks all sorts of assumptions that other parts of the code have about that `forward()` method.

ZhaofengWu:
> This does not work for T5 at all 😭. I'm no longer sure this approach of patching the model will work. The huggingface generation code makes calls into the middle of their model, instead of always going through the `forward()` method. So patching `forward()` doesn't work. 

Is this problematic for generation only?

> And patching `forward()` of an internal method breaks all sorts of assumptions that other parts of the code have about that `forward()` method.

This is what I was worrying about above.

dirkgr:
Copying from Slack:

I can patch just the encoder for T5. Then the soft prompt has the opportunity to change how the rest of the prompt is encoded. But the encoded soft tokens are not part of the encoder output, and cannot be attended to by the decoder. @ZhaofengWu, is that important?

dirkgr:
Just to resolve this chain of comments: I made it work with T5.

tango/integrations/transformers/soft_prompt.py
```diff
@@ -0,0 +1,221 @@
...
+        torch.empty(
+            1,
+            prompt_length,
+            original_embedding.embedding_dim,
+            dtype=original_embedding.weight.dtype,
+            device=original_embedding.weight.device,
+        )
+    )
+    r = random.Random(random_seed)
+    indices = torch.tensor(r.sample(range(5000), prompt_length))
```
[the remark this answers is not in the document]
dirkgr: That sure is a little weird. Maybe it should sample from the entire original embedding.

CHANGELOG.md
```diff
@@ -262,6 +262,7 @@ instead of `ModuleNotFound`.
 - Added the "-n/--name" option to `tango run`. This option allows the user to give the run an arbitrary name.
 - Added a convenience property `.workspace` to `Step` class that can be called from a step's `.run()` method to get the current `Workspace` being used.
 - Gave `FromParams` objects (which includes all `Registrable` objects) the ability to version themselves.
+- Added the `transformers::with_soft_prompt` integration, to make soft-prompted prefix transformers easy.
```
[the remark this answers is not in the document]
dirkgr: 52b876bdc37881a10363551890302bc36731d47e

tango/integrations/transformers/soft_prompt.py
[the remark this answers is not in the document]
dirkgr: 59f57b06cd7aacbc9d31268ff98290fc4fd0343c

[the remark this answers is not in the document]
ZhaofengWu: It was originally used in https://arxiv.org/abs/2104.08691 and subsequently other papers such as https://arxiv.org/abs/2108.04106 and of course ours. The idea is to only use the representation of the top-5000 tokens.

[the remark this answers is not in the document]
ZhaofengWu: I'd keep 5000 or at least have some flag to control this.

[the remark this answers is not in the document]
dirkgr: I made it configurable, with a default of 5000.

dirkgr:
This is ready for another review.

tango/integrations/transformers/soft_prompt.py
```diff
@@ -0,0 +1,248 @@
...
+        inputs_embeds = kwargs.get("inputs_embeds", inputs_embeds)
+        if inputs_embeds is not None:
+            kwargs["inputs_embeds"] = torch.cat(
+                [prompt_embedding.expand(inputs_embeds.size(0), -1, -1), inputs_embeds], dim=1
+            )
+
+        patch_tensor(kwargs, "labels")
+        patch_tensor(kwargs, "attention_mask", 1)
+        patch_tensor(kwargs, "token_type_ids")
+        patch_tensor_with_indices(kwargs, "position_ids", prompt_length)
```
AkshitaB, line 147: So, if the position ids are originally `[0, 1, 2, 3, 4]`, they will now be `[0, 1, 2, .. prompt_len-1, 0, 1, 2, 3, 4]` ?

dirkgr: Yes, that's right. I could see it going the other way, but I think it's important that the output does not change in the case where the soft prompt is configured to do nothing. Also, if we offset the position ids, we would decrease the max length that the model can handle, which is uncomfortable.

AkshitaB (approved):

PR #232, opened 2022-03-15
Add Weights & Baises remote workspace

### LINKED ISSUE #125: Remote workspace
This is about a workspace implementation that stores the cache and assorted run information remotely. It's an implementation of the `Workspace` class that keeps the data in the cloud. We want this because we want to see the same workspace from servers at Emerald Landing, Cirrascale, GCE, and also from our work laptops.

I think the two leading candidates for storing step results are wandb and GCS. In my mind Google is going to be a more stable, conservative choice, but wandb has a ton of cool integrations and visualizations. If wandb is equally easy to set up, equally fast, and equally scalable (1TB for a model checkpoint?), I'm not sure we still need something that saves the data in GCS.

In `LocalWorkspace` I went out of my way to store run information and step information directly in the file system. It's nice to have the same system for both step results and metadata, because that's one less system. For a remote workspace, it might not scale to access a ton of tiny files in GCS though, so maybe it's time to bite the bullet and put it into a DB. If you choose to go that route, please give me a chance to convince you of the merits of writing SQL queries directly in Python code before you start using a persistence framework.

The remote workspace will need a local cache for step results, both for performance, and also because we want to memory map some of those files. This should be easy, since all the step results are immutable.

### LINKED ISSUE #235: MulticoreExecutor doesn't preserve custom workspace
## Checklist

- [x] I have verified that the issue exists against the `main` branch.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/tango/blob/main/CONTRIBUTING.md#bug-reports-and-feature-requests) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/tango/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/tango/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/tango/blob/main/CHANGELOG.md) and the [commit log](https://github.com/allenai/tango/commits/main) to find out if the bug was already fixed in the main branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

Apologies in advance if this is intended behavior.
When defining a custom workspace, MulticoreExecutor will alway uses the default workspace for parallel steps due to the line here:
https://github.com/allenai/tango/blob/410faebaf429a48d02c2690ab2637d01b2b5fac0/tango/executors/multicore_executor.py#L226
To correct it, we would need to prepend the registered name to `self.workspace.dir` or store the unparsed url

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
```

</p>
</details>

## Related issues or possible duplicates
- https://github.com/allenai/tango/pull/232 (I looked over this PR and I don't think it resolves it, but I could be wrong)

## Environment

OS: Linux

Python version: 3.8.12

<details>
<summary><b>Output of <code>pip freeze</code>:</b></summary>
<p>

```
ai2-tango @ git+https://github.com/allenai/tango.git@410faebaf429a48d02c2690ab2637d01b2b5fac0
```

</p>
</details>

## Steps to reproduce

- Create a new workspace that inherits from LocalWorkspace and register it.
- Run and check the type of the workspace when `called_by_executor == True`

<details>
<summary><b>Example source:</b></summary>
<p>

```
from tango import Workspace, Step
from tango.common import PathOrStr
from tango.workspaces.local_workspace import LocalWorkspace, T, logger


@Workspace.register("test")
class TestWorkspace(LocalWorkspace):
    def __init__(self, directory: PathOrStr):
        logger.info("Initializing test workspace", dir=directory)
        super().__init__(directory)

    def step_starting(self, step: Step) -> None:
        logger.info("This will never be called")
        return super().step_starting(step)

    def step_finished(self, step: Step, result: T) -> T:
        logger.info("This will never be called")
        result = super().step_finished(step, result)
        return result
```

</p>
</details>

### LINKED ISSUE #278: Problems running the Euler example
I'm also a bit confused, because when I run the following:

```
tango run euler_general.jsonnet -i complex_arithmetic -d workspace
```

I would expect my work to be in the folder workspace.  But when I ls this folder it's empty and instead I see my run directly in runs.

```
$ ls workspace/runs/
(empty)
$ ls runs
hot-marlin/   viable-goose/
```

I also don't seem to have step_result_for_run that the CLI output suggested.

```
>>> from tango import Workspace
>>> workspace = Workspace.from_url('.')
>>> workspace.step_result_for_run
  File "<stdin>", line 1, in <module>
AttributeError: 'LocalWorkspace' object has no attribute 'step_result_for_run'
```

### LINKED ISSUE #281: Deprecation warnings not being logged

epwalsh:
Closes #125, fixes #235, fixes #278, fixes #281

- Adds a Weights & Biases remote workspace implementation: `WandbWorkspace`.
- Improves the `WandbTrainCallback` in several ways, making it work seamlessly with `WandbWorkspace`.
- Adds `Workspace.url` property (must be implemented by subclasses), which returns a URL suitable to instantiate the same workspace using `Workspace.from_url()`.
- Adds `Workspace.step_result_for_run()` method, which gives you the result of a particular step from a particular run, given the name of the step and the name of the run (fixes one of the issues brought up in #278).
- Fixes a bug with `LocalWorkspace.from_parsed_url()`, which resulted in the workspace using the wrong directory (fixes the other issue from #278).
- Ensures deprecation warnings are displayed from the CLI (#281).

## Tips for reviewers

This pull request has touched a lot of files, so here are some tips for reviewing efficiently.

1. Start by inspecting the first commit ([3eaa483](https://github.com/allenai/tango/pull/232/commits/3eaa4831290d8fd5537612e3ba4524b000ea409b)), which lays the groundwork for the subsequent commits. This commit doesn't touch anything related to Weights & Biases, but it makes some key fixes to the `Step` class that are needed for the `WandbWorkspace`, as well some fixes and additions to `StepCache`, `Workspace`, and `StepInfo`.

2. If you're satisfied with the first commit, move on to the second commit ([bfd6a10](https://github.com/allenai/tango/pull/232/commits/bfd6a10ceee56316a7901f147fd68e8d947f2342)), which actually adds the `WandbWorkspace`. I'd suggest paying close attention to the logic in `WandbWorkspace` and `WandbStepCache`, in particular.

3. Next, I recommend creating your own workspace on Weights & Biases, and running some mock tango experiments with the `WandbWorkspace`. For example:

    ```
    tango run test_fixtures/experiment/hello_world.jsonnet -i test_fixtures/package -w wandb://allennlp/my-tango-testing
    ```

    Browse the results on the W&B dashboard and see if you can make sense of them. If you can't, then maybe this is useless 🤨
    
4. Lastly, I recommend reviewing the documentation additions. For that, just build the docs locally and inspect the page for the `wandb` integration.

## Before submitting

## After submitting

## Review items to complete

- [x] Change naming of step artifacts to not use step name (https://github.com/allenai/tango/pull/232#discussion_r847793548)
- [x] Use local timezone in timestamps, or at least when displaying times (https://github.com/allenai/tango/pull/232#discussion_r847780312)
- [x] Add test with multi executor (https://github.com/allenai/tango/pull/232#discussion_r845668692)
- [x] Make sure web UI works fine with a W&B workspace
- [ ] Try running two workspaces in two separate threads


tango/integrations/wandb/step_cache.py
```diff
@@ -0,0 +1,227 @@
...
+            key = step.unique_id
+
+            # First check if we have a copy in memory.
+            if key in self.strong_cache:
+                return True
+            if key in self.weak_cache:
+                return True
+
+            # Then check if we have a copy on disk in our cache directory.
+            with self._acquire_step_lock_file(step, read_only_ok=True):
```
AkshitaB, line 127: Is there a possibility of the disk cache dir and W&B being out of sync?

tests/integrations/wandb/workspace_test.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @pytest.mark.parametrize(
+        "distributed",
+        [
+            pytest.param(True, id="distributed"),
+            pytest.param(False, id="single-device"),
+        ],
+    )
+    def test_with_wandb_train_callback(self, distributed: bool):
+        self.run(
```
AkshitaB, line 149: We should also have a test with `multicore = True` in `run()`.

tango/__main__.py
```diff
@@ -727,7 +727,10 @@ def log_step_summary(step):
                 click_logger.info(
                     click.style("\N{check mark} The output for ", fg="green")
                     + click.style(f'"{step.name}"', bold=True, fg="green")
-                    + click.style(" is in ", fg="green")
+                    + click.style(
+                        " is in " if Path(info.result_location).exists() else " is at ",
```
dirkgr, line 731: love the attention to detail

tango/integrations/wandb/workspace.py
```diff
@@ -0,0 +1,429 @@
...
+            wandb.init(
+                name=step_info.step_name,
+                job_type=RunKind.STEP.value,
+                group=step.unique_id,
+                dir=str(work_dir),
+                entity=self.entity,
+                project=self.project,
+                # For cacheable steps we can just use the step's unique ID as the W&B run ID,
+                # but not for uncacheable steps since those might be ran more than once, and
+                # and will need a unique W&B run ID each time.
```
dirkgr, line 164: Do uncacheable steps need a wandb run ID at all? They are usually just folded into the step that has them as a dependency ... except for uncacheable leaf steps.

tango/integrations/wandb/workspace.py
```diff
@@ -0,0 +1,429 @@
...
+        if wandb.run is None:
+            raise RuntimeError(
+                f"{self.__class__.__name__}.step_failed() called outside of a W&B run. "
+                f"Did you forget to call {self.__class__.__name__}.step_starting() first?"
+            )
+
+        step_info = self._running_step_info[step.unique_id]
+
+        try:
+            # Update StepInfo, marking the step as finished.
```
dirkgr, line 255: Marking it as failed, surely

dirkgr (changes requested):
First pass. I haven't tried to run it yet.

tango/integrations/wandb/workspace.py
epwalsh: Not necessarily, but I think it's better this way as it results in a complete execution graph in W&B.

tango/integrations/torch/eval_callback.py
```diff
@@ -19,24 +20,24 @@ class EvalCallback(Registrable):
...
 
-    Attributes
-    ----------
-    step_id : :class:`str`
-    work_dir : :class:`~pathlib.Path`
-    model : :class:`Model`
-    dataset_dict : :class:`tango.common.DatasetDictBase`
-    dataloader : :class:`DataLoader`
-
+    :ivar Workspace workspace: The tango workspace being used.
```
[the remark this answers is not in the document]
epwalsh: "Instance variable"

tango/integrations/wandb/workspace.py
```diff
@@ -0,0 +1,429 @@
...
+    @property
+    def url(self) -> str:
+        return f"wandb://{self.entity}/{self.project}"
+
+    @classmethod
+    def from_parsed_url(cls, parsed_url: ParseResult) -> Workspace:
+        entity = parsed_url.netloc
+        project = parsed_url.path
+        if project and project.startswith("/"):
+            project = project[1:]
```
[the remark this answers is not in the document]
epwalsh: 159b0e7

tango/integrations/wandb/workspace.py
epwalsh: 159b0e7

tango/step_caches/local_step_cache.py
```diff
@@ -68,44 +90,52 @@ def _get_from_cache(self, key: str) -> Optional[Any]:
         except KeyError:
             return None
 
+    def _metadata_path(self, step_or_unique_id: Union[Step, StepInfo, str]) -> Path:
+        return self.step_dir(step_or_unique_id) / self.METADATA_FILE_NAME
+
     def __contains__(self, step: object) -> bool:
-        if isinstance(step, Step) and step.cache_results:
+        if isinstance(step, (Step, StepInfo)):
```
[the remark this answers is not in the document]
epwalsh: No good reason. It's back now: 5b5e51f

tango/integrations/wandb/step_cache.py
```diff
@@ -0,0 +1,227 @@
...
+    def __len__(self) -> int:
+        completed_cacheable_step_runs = self.wandb_client.runs(
+            f"{self.entity}/{self.project}",
+            filters={
+                "config.job_type": "step",
+                "config.cacheable": True,
+                "state": "finished",
+            },
+        )
+        return len(list(completed_cacheable_step_runs))
```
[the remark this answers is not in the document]
epwalsh: a8268a6

tango/integrations/wandb/step_cache.py
```diff
@@ -0,0 +1,227 @@
...
+
+        if wandb.run is None:
+            raise RuntimeError("Can only add results to the WandbStepCache within a W&B run")
+
+        with self._acquire_step_lock_file(step):
+            # Overwrite if necessary. If `value` ends up being the same as the value that we're
+            # overwriting, the checksum should also be the same, and so W&B won't actually upload
+            # any new files.
+            if self.step_dir(step).is_dir():
+                shutil.rmtree(self.step_dir(step), ignore_errors=True)
```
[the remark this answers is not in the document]
epwalsh: a8268a6

tango/integrations/wandb/step_cache.py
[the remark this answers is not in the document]
epwalsh: Another process could be in the process of writing the artifact to that directory, in which case we want to wait until that process is finished before saying "yeup, we can give you that step's artifact"

tango/integrations/wandb/step_cache.py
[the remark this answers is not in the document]
epwalsh: It shouldn't be that expensive. It's basically just making a MongoDB query. I'm assuming they at least have an index on the "state" field.

tango/integrations/wandb/step_cache.py
```diff
@@ -0,0 +1,227 @@
...
+            return step.step_name if step.step_name is not None else step.unique_id
+
+    def get_step_result_artifact(
+        self, step: Union[Step, StepInfo]
+    ) -> Optional[wandb.apis.public.Artifact]:
+        try:
+            return self.wandb_client.artifact(
+                f"{self.entity}/{self.project}/{self._step_artifact_name(step)}:{step.unique_id}",
+                type=ArtifactKind.STEP_RESULT.value,
+            )
```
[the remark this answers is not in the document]
epwalsh: Hmm good point. I was going for readability but overlooked that. I'll have to change this.

tango/step_caches/local_step_cache.py
```diff
@@ -30,18 +33,37 @@ class LocalStepCache(StepCache):
...
+        self.strong_cache = collections.OrderedDict()
+
+    def __getstate__(self):
+        """
+        We override `__getstate__()` to customize how instances of this class are pickled
+        since we don't want to persist values in the weak and strong in-memory caches
+        during pickling. And WeakValueDictionary can't be pickled anyway.
+        """
+        out = {k: v for k, v in self.__dict__.items() if k not in {"weak_cache", "strong_cache"}}
+        return out
```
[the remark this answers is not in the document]
epwalsh: Unfortunately yes. For example, during distributed training we pass the workspace to worker processes: https://github.com/allenai/tango/pull/232/files#diff-72a1daf1ee839673aee1e5e2b0b2a65111e0e72bbaeaa07a5321ee88882156caR315

tests/integrations/wandb/workspace_test.py
epwalsh: Adding that to the TODO list 👍

tango/integrations/wandb/workspace.py
```diff
@@ -0,0 +1,429 @@
...
+        out = super().__getstate__()
+        out["locks"] = {}
+        return out
+
+    @property
+    def wandb_client(self) -> wandb.Api:
+        overrides = {"project": self.project}
+        if self._entity is not None:
+            overrides["entity"] = self._entity
+        return wandb.Api(overrides=overrides)
```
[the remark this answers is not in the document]
epwalsh: I think it's okay, but here's a better solution: a39c5ec

tango/integrations/wandb/step_cache.py
```diff
@@ -0,0 +1,227 @@
...
+        self.entity = entity
+
+    @property
+    def wandb_client(self) -> wandb.Api:
+        return wandb.Api(overrides={"entity": self.entity, "project": self.project})
+
+    def _acquire_step_lock_file(self, step: Union[Step, StepInfo], read_only_ok: bool = False):
+        return FileLock(
+            self.step_dir(step).with_suffix(".lock"), read_only_ok=read_only_ok
+        ).acquire_with_updates(desc=f"acquiring step cache lock for '{step.unique_id}'")
```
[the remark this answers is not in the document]
epwalsh: I think the reason `LocalWorkspace` does that is because it's built so that it can be running multiple steps at the same time from the same `LocalWorkspace` instance. But that's not possible at the moment with `WandbWorkspace`, since you can only have one W&B run in a process at once.

tango/workspaces/memory_workspace.py
```diff
@@ -103,25 +94,18 @@ def step_failed(self, step: Step, e: BaseException) -> None:
...
-                step.__class__.__name__,
-                step.VERSION,
-                {dep.unique_id for dep in step.dependencies},
-                step.cache_results,
-            )
+            step_info = StepInfo.new_from_step(step)
             self.unique_id_to_info[step.unique_id] = step_info
             steps[step.unique_id] = step_info
-        run = Run(name, steps, datetime.now())
+        run = Run(name, steps, utc_now_datetime())
```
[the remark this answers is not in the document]
epwalsh: I'm a huge fan of UTC, and there are some problems with using local timezones, e.g. see https://pythonhosted.org/pytz/#problems-with-localtime. 

pytz recommends always storing dates in UTC, and only converting to local times when displaying dates. Maybe we could take that approach.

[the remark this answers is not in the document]
epwalsh: Putting this here for myself, the first comment in [this SO post](https://stackoverflow.com/a/39079819/4151392) shows how to get the local timezone effectively for Python >=3.6:

```python
datetime.datetime.now().astimezone().tzinfo
```

tango/integrations/wandb/step_cache.py
epwalsh: Not for an individual step, unless someone messes with the contents of the directory. But that shouldn't happen. W&B artifacts are versioned, and each local step cache directory is tied to a specified version of the artifact on W&B. A version of an artifact is immutable AFAIK.

tango/workspaces/local_workspace.py
```diff
@@ -72,7 +74,7 @@ def __init__(self, dir: PathOrStr):
             with SqliteDict(self.step_info_file) as d:
                 for stepinfo_file in self.cache.dir.glob("*/stepinfo.dill"):
                     with stepinfo_file.open("rb") as f:
-                        stepinfo = StepInfo.deserialize(f.read())
+                        stepinfo = dill.load(f)
```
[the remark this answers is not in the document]
epwalsh: It's being used by the `WandbWorkspace`. I think it would be better to store step info here as JSON as well, but that would be a breaking change and I wanted to limit the scope of this PR as best I could.

tango/integrations/wandb/step_cache.py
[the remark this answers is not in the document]
epwalsh: af36134

tests/integrations/wandb/workspace_test.py
epwalsh: Done: 143f23c

tango/integrations/wandb/workspace.py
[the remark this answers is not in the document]
epwalsh: 🤦‍♂️ turns out my original implementation was necessary, caching it via https://github.com/allenai/tango/commit/a39c5ece0f8d1605ff9993cc1da57de23f7be529 actually breaks things since `wandb.Api` does some internal caching of HTTP requests to the server. I finally figured out that this was causing the bug with step state being out-of-date. FYI @AkshitaB 

Reverted in 5894941

tango/workspaces/memory_workspace.py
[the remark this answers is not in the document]
dirkgr: > storing dates in UTC, and only converting to local times when displaying dates

Yes, definitely.

tango/integrations/wandb/torch_train_callback.py
```diff
@@ -15,36 +18,47 @@ class WandbTrainCallback(TrainCallback):
     the :class:`~tango.integrations.torch.TorchTrainStep` that logs training and
     validation metrics to W&B.
 
+    This can be used with any :class:`~tango.workspace.Workspace` implementation,
+    including :class:`WandbWorkspace`.
```
dirkgr, line 22: All the customization you put in here, do we also need that for the lightning integration?

tango/integrations/wandb/workspace.py
```diff
@@ -0,0 +1,427 @@
...
+        return self.cache
+
+    @property
+    def wandb_project_url(self) -> str:
+        """
+        The URL of the W&B project this workspace uses.
+        """
+        app_url = self.wandb_client.client.app_url
+        if app_url.endswith("/"):
+            app_url = app_url[:-1]
```
dirkgr, line 98:
```suggestion
        app_url = app_url.rstrip("/")
```
This will also handle multiple trailing slashes.

tango/step.py
```diff
@@ -354,18 +359,31 @@ def _run_with_work_dir(self, workspace: "Workspace", **kwargs) -> T:
             self.work_dir_for_run = Path(dir_for_cleanup.name)
 
         try:
-            if self.cache_results:
-                workspace.step_starting(self)
```
dirkgr, line 358: We used to have a thing where workspaces never get bothered with uncacheable results. It's also OK if we say, workspaces must handle uncacheable results at all times. But we have to be consistent about it.

tango/step.py
```diff
@@ -354,18 +359,31 @@ def _run_with_work_dir(self, workspace: "Workspace", **kwargs) -> T:
...
             try:
                 result = self.run(**kwargs)
-                if self.cache_results:
-                    result = workspace.step_finished(self, result)
+                result = workspace.step_finished(self, result)
+                cli_logger.info(f'[green]\N{check mark} Finished step [bold]"{self.name}"[/][/]')
                 return result
             except BaseException as e:
                 # TODO (epwalsh): do we want to handle KeyboardInterrupts differently?
                 # Maybe have a `workspace.step_interrupted()` method?
```
dirkgr, line 385: No. Why would we?

tango/step_cache.py
```diff
@@ -1,11 +1,13 @@
...
+from typing import Any, TypeVar, Union
 
-from tango.common.from_params import FromParams
-from tango.common.registrable import Registrable
-from tango.step import Step
+from .common.from_params import FromParams
+from .common.registrable import Registrable
+from .format import Format
+from .step import Step
+from .step_info import StepInfo
```
dirkgr, line 10: I thought we don't like relative imports? Don't they cause some sort of trouble with `Registrable`?

tango/step_caches/local_step_cache.py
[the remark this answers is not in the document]
dirkgr: I think it is in the nature of this class that the only thing we actually want to store is the directory that contains the cache, right? Because the cache guarantees that it will look the same from every process that uses that directory, by definition the directory must be enough to pickle the cache. So I'm thinking, let's just return the directory from `__getstate__`.

tango/step_caches/local_step_cache.py
```diff
@@ -68,44 +90,54 @@ def _get_from_cache(self, key: str) -> Optional[Any]:
...
             self._add_to_cache(key, result)
         return result
 
     def __setitem__(self, step: Step, value: Any) -> None:
         if not step.cache_results:
-            logger.warning("Tried to cache step %s despite being marked as uncacheable.", step.name)
+            warnings.warn(
+                f"Tried to cache step '{step.name}' despite being marked as uncacheable",
+                UserWarning,
+            )
```
dirkgr, line 127: The `logger` construction has the advantage that it won't do the string substitution unless the string is actually needed (i.e., warnings are not silenced). A `f""` string is always created first, before we find out whether the string is needed.

tango/step_caches/local_step_cache.py
```diff
@@ -117,17 +149,27 @@ def __setitem__(self, step: Step, value: Any) -> None:
...
+            if not cacheable:
+                class_name = (
+                    step_or_unique_id.__class__.__name__
+                    if isinstance(step_or_unique_id, Step)
+                    else step_or_unique_id.step_class_name
+                )
                 raise RuntimeError(
-                    f"Uncacheable steps (like '{step_or_unique_id.name}') don't have step directories."
+                    f"Uncacheable steps (like '{class_name}') don't have step directories."
```
dirkgr, line 172: I'm OK with this, but in principle, one instance of a step class could be cacheable, and another could be not cacheable.

dirkgr (commented):
Have you tried running the Tango Web UI against this workspace? It often exposes some unusual access patterns.

tango/workspace.py
```diff
@@ -228,6 +85,23 @@ class Workspace(Registrable):
...
+    @property
+    def url(self) -> str:
+        """
+        Get a URL for the workspace that can be used to instantiate the same workspace
+        using :meth:`.from_url()`.
+        """
+        raise NotImplementedError
+
     @classmethod
     def from_url(cls, url: str) -> "Workspace":
```
dirkgr, line 106: The fact that this takes URLs as well as local paths seems worth mentioning in the docs.

tango/step_caches/local_step_cache.py
epwalsh: I'll bet any potential perf benefits of `logger.warning` vs `warnings.warn` are completely negligible in this case, and in my opinion usage warnings from a library should always be emitted via the `warnings` module so that warnings will propagate even if logging isn't configured.

tango/workspace.py
epwalsh: In addition to this?
https://github.com/allenai/tango/blob/9e6c3ee3caf02f5b3f011014377359473a8f3d10/tango/workspace.py#L108

tango/step_cache.py
epwalsh: I don't not like (explicit) relative imports. I prefer them actually, and I'm not aware of any issues with `Registrable`.

tango/step.py
epwalsh: To distinguish between steps that failed and steps that were interrupted. But I don't care enough to argue. 5f1ac38

tango/integrations/wandb/torch_train_callback.py
epwalsh: I haven't tried to use PTL's `WandbLogger`, but looking at [its source code](https://pytorch-lightning.readthedocs.io/en/stable/_modules/pytorch_lightning/loggers/wandb.html#WandbLogger) it seems like it would work okay with the `WandbWorkspace` because it reuses any existing W&B run when there is one.

tango/integrations/wandb/step_cache.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # Log/persist the artifact to W&B.
+        artifact.save()
+        artifact.wait()
+
+        # Add an alias for the step's unique ID.
+        # Only after we've logged the artifact can we add an alias.
+        artifact.aliases.append(step.unique_id)
+        artifact.save()
+        artifact.wait()
```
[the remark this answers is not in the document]
epwalsh: Yes, I believe so.

epwalsh:
> Have you tried running the Tango Web UI against this workspace?

Yes, had to make one fix: a8ce1b0

tango/workspace.py
dirkgr: I checked the code to make sure that `"/tmp/workspace"` also still works. It wasn't clear to me from the docs.

tango/step_cache.py
dirkgr: I also prefer relative imports, but working on the AllenNLP codebase educated me out of it. I'm fine, happy even, to try again in Tango.

dirkgr (approved):
Only thing left is that one doc change. And even if that doesn't make it, this is a huge step for Tango!

epwalsh:
Done: 1c2d0b4

dirkgr:
Is this gtg then?

PR #233, opened 2022-03-15
Logging improvements

epwalsh:
Changes proposed:
- Renames `click_logger` to `cli_logger` and uses the [rich](https://github.com/Textualize/rich) logging `Handler` for prettier logs and tracebacks that are much easier to read:

    image

- That also allows us to uses rich's inline markup syntax instead of having to wrap text in `click.style()` everywhere.
    So code like this:
    
    image

    Can be refactored into this:
    
    image

- Changes the behavior of `cli_logger` based on the log level. As before, ALL messages (regardless of their level) will be emitted from the `cli_logger` (as long as the CLI logging has been enabled). But now any message with a level at or above the set log level will be displayed the same as any other log message from a different logger. This results in more consistent output formatting, especially when log level is set at `INFO`.

    Here are some examples from a step that emits message both from `self.logger` and `cli_logger`. At `ERROR` log level:
    
    image
    
    At `WARNING` log level:
    
    image

    At `INFO` log level:

    image

    And at `DEBUG` log level:

    image

- Rich can even print code examples with syntax highlighting, which let's us print out useful snippets like this:

    image


dirkgr (approved):
This is very cool, and we should have it.

There is one thing I worry about with all the colors and prettiness: It undercuts our message that Tango is lightweight and easy to adopt. It's a perception thing, not a technical reality. But when we give demos with pretty colors (and how could we not?), people will automatically put Tango into a category, and it's a category that we might not want to be in.

PR #241, opened 2022-03-23
Set `$TEMP`

dirkgr:


epwalsh (approved):

dirkgr:
This breaks Python's multiprocessing, the bane of my existence. Reverted in bbdeb6f.

Issue #244, opened 2022-03-28
Some dependencies don't show up as a step's dependencies when they should


dirkgr:
I think this happens when a dependency is not directly in a step's `kwargs`, but is somewhere nested in there.

dirkgr:
This was fixed in #249.

PR #249, opened 2022-03-30
Fix tracking dependencies between steps

dirkgr:
🙃


dirkgr:
Fixes #244

AkshitaB (changes requested):
We should probably add a test case for the bug you noticed.

AkshitaB (approved):

PR #252, opened 2022-03-30
Logging part 2: denoising run logging and making Dirk happy

epwalsh:
Fixes some issues brought up internally with log output being too noisy.

- Don't print the traceback for "Run finished with errors". It's not useful, and takes attention away from the step tracebacks that actually matter.
- Prints a summary at the end of the run, and a single example of how to get cached results using a workspace, but only if the run passed.
  
  Example from a failed run:
  image

  Example from a successful run:  
  image

- Increases the default screen width when no terminal is detected (like if being run by a debugger).

## Before submitting

## After submitting


tango/common/logging.py
```diff
@@ -608,12 +645,15 @@ def file_handler(filepath: PathOrStr) -> ContextManager[None]:
         force_terminal=False,
         width=TANGO_CONSOLE_WIDTH,
     )
+    if TANGO_CONSOLE_WIDTH is None and not console.is_terminal:
+        console.width = 160
```
dirkgr, line 649: 🔥

dirkgr (approved):

PR #257, opened 2022-04-06
Bug fix with failing steps

AkshitaB:

- We now check for failed steps in `recursive_dependencies` rather than `dependencies`. This allows the `MulticoreExecutor` to exit successfully instead of hanging in cases like `A --> B --> C` where `A` fails, but the executor is still hoping to run `C` at some point.

## Before submitting

## After submitting


tango/step.py
```diff
@@ -115,6 +115,7 @@ def __init__(
         cache_results: Optional[bool] = None,
         step_format: Optional[Format] = None,
         step_config: Optional[Dict[str, Any]] = None,
+        step_unique_id: Optional[str] = None,
```
dirkgr, line 118: This needs documentation. And maybe let's call it `step_unique_id_override` to make it a little clearer what the relationship is between this value and the step's unique ID.

dirkgr (changes requested):

dirkgr (approved):

PR #258, opened 2022-04-07
use beaker-py to submit GPU tests

epwalsh:
This allows us to simplify the GitHub Actions config file and dynamically select the cluster to run on based off of GPU availability. That means we can run on on-prem clusters when they have free GPUs, otherwise we fall back to the scalable cloud cluster.


.github/workflows/main.yml
```diff
@@ -15,91 +15,102 @@ on:
...
+    name: ${{ matrix.task.name }}
     runs-on: [ubuntu-latest]
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
         python: ['3.7']
         task:
-          - name: Lint
+          - name: python 3.7 - Lint
```
dirkgr, line 36: Isn't it bad to hard-code the version number like this?

scripts/beaker_submit_gpu_tests.py
```diff
@@ -0,0 +1,102 @@
...
+    beaker,
+)
+
+
+class TermInterrupt(Exception):
+    pass
+
+
+def handle_sigterm(sig, frame):
+    raise TermInterrupt
```
dirkgr, line 38: What's this for?

scripts/beaker_submit_gpu_tests.py
```diff
@@ -0,0 +1,102 @@
...
+    # but we'll also check to see if we can find an on-prem cluster with enough
+    # free GPUs.
+    beaker_cluster = BEAKER_CLOUD_CLUSTER
+    for on_prem_cluster in BEAKER_ON_PREM_CLUSTERS:
+        for node_util in beaker.cluster.utilization(on_prem_cluster):
+            if node_util.free.gpu_count is not None and node_util.free.gpu_count >= 2:
+                beaker_cluster = on_prem_cluster
+                print(
+                    f"- Found on-prem cluster with enough free GPUs ({on_prem_cluster})", end="\n\n"
+                )
```
dirkgr, line 56: This ability should really be part of Beaker.

scripts/beaker_submit_gpu_tests.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def handle_sigterm(sig, frame):
+    raise TermInterrupt
+
+
+def main(name: str, commit_sha: str):
+    signal.signal(signal.SIGTERM, handle_sigterm)
+
+    print(f"- Authenticated as {beaker.account.name}", end="\n\n")
```
dirkgr, line 44: Print instead of logging?

scripts/beaker_submit_gpu_tests.py
```diff
@@ -0,0 +1,102 @@
...
+        print("- Waiting for job to finish...", end="\n\n")
+        experiment = beaker.experiment.await_all(experiment, timeout=20 * 60)
+
+        print("- Pulling logs...", end="\n\n")
+        logs = "".join([line.decode() for line in beaker.experiment.logs(experiment)])
+        rich.get_console().rule("Logs")
+        print(logs)
+
+        sys.exit(experiment.jobs[0].status.exit_code)
+    except (KeyboardInterrupt, TermInterrupt):
```
dirkgr, line 96: Ah, I get it now.

dirkgr (approved):

scripts/beaker_submit_gpu_tests.py
epwalsh: GH Actions sends a SIGTERM when a job is cancelled, which Python doesn't catch by default.

.github/workflows/main.yml
epwalsh: It's not ideal, but the GPU tests shouldn't have the Python version in their name. That's the one exception, which is why I had to change this.

Why shouldn't the GPU tests have the Python version in their name? Because the Python version here is not relevant to the GPU tests since the GPU tests are ran in a Docker container which may be using a different Python version.

scripts/beaker_submit_gpu_tests.py
epwalsh: Yeup, they're working on it.

scripts/beaker_submit_gpu_tests.py
epwalsh: Hm, maybe I'll switch to logging.

epwalsh: [d8f4c16](https://github.com/allenai/tango/pull/258/commits/d8f4c1679ea02ef4d2ffa00927bc201610d97181)

PR #289, opened 2022-05-18
Makes sure the `VERSION` parameter of classes is respected even when we construct them inside of a `Lazy` object.

dirkgr:


tango/common/lazy.py
```diff
@@ -85,3 +86,54 @@ def construct(self, **kwargs) -> T:
...
+            params = copy.deepcopy(self._params)
+            if params is None:
+                params = Params({})
+            elif isinstance(params, str):
+                params = Params({"type": params})
+            elif isinstance(params, dict):
+                params = Params(params)
+
+            if not isinstance(params, Params):
+                return None
```
epwalsh, line 104: This doesn't seem right, did you mean to have this here?

tango/common/registrable.py
```diff
@@ -184,7 +184,12 @@ def search_modules(cls: Type[_RegistrableT], name: str):
         def try_import(module):
             try:
                 import_module_and_submodules(module)
-            except (ModuleNotFoundError, ImportError):
+            except ModuleNotFoundError as e:
+                if e.name == module:
+                    pass
+                else:
+                    raise
+            except ImportError:
```
epwalsh, line 192: This is causing some test failures

tango/common/lazy.py
dirkgr: Yes. When you return `None` from `det_hash()`, it will fall back to just hashing the object itself as if you were not using `CustomDetHash` at all.

epwalsh, line 104: Gotcha. How about:
```suggestion
                params = Params(params)
            elif not isinstance(params, Params):
                return None
```

I think that's a little more clear. Just a suggestion though.

tango/common/registrable.py
dirkgr: Did we rely on `ImportError` being silently swallowed to make tests work?

epwalsh: We need this because we might try to import an integration that isn't installed, i.e. an integration that is missing some dependencies, like torch.

epwalsh (approved):

PR #296, opened 2022-05-25
Add Beaker integration

epwalsh:
Adds a `BeakerWorkspace`.

Test this out with:

```bash
tango run test_fixtures/experiment/hello_world.jsonnet -i test_fixtures/package -w beaker://ai2/tango-beaker-testing
```

This is a prototype at the moment so I'd like people to try it out. There's still a few enhancements I have planned:
- [x] Capture and upload logs from each run. To implement this, we'd need to implement `BeakerWorkspace.capture_logs_for_run()` similar to how `LocalWorkspace` does it, except maybe we could use a file handler that pipes to a temporary file instead of a "permanent" file. Then we upload that file to the run dataset. So we'd have to postpone committing the run dataset until `__exit__` on the `.capture_logs_for_run()` context manager is called. (90f837f)
- [x] Utilize a thread pool in situations where we could make a bunch of requests concurrently, such as in `BeakerWorkspace.registered_runs()` and `BeakerWorkspace._get_run_from_dataset()`. (49242aa)
- [x] In addition to the local file lock this workspace grabs on `.step_starting()`, we could also grab a "lock" on Beaker. This lock could be implemented using Beaker datasets. E.g.: a process creates a dataset with a deterministic name, and holds that dataset as the lock until it releases it by deleting the dataset. If any other process tried to grab that same lock dataset, they would get a `DatasetConflict` error from Beaker. (1bc7cc3)

### DIFF
--- a/tango/integrations/beaker/step_cache.py
+++ b/tango/integrations/beaker/step_cache.py
@@ -167,12 +167,10 @@ class BeakerStepCache(LocalStepCache):
         self._add_to_cache(step.unique_id, value)
 
     def __len__(self) -> int:
-        return len(
-            [
-                ds
+        return sum(
+                1
                 for ds in self.beaker.workspace.datasets(
                     uncommitted=False, match=Constants.STEP_DATASET_PREFIX
                 )
                 if ds.name is not None and ds.name.startswith(Constants.STEP_DATASET_PREFIX)
-            ]
         )
--- a/tango/integrations/beaker/step_cache.py
+++ b/tango/integrations/beaker/step_cache.py
@@ -1,0 +1,178 @@
+import logging
+import os
+import shutil
+import tempfile
+from pathlib import Path
+from typing import Any, Optional, Union
+
+from beaker import Beaker, Dataset, DatasetConflict, DatasetNotFound, DatasetWriteError
+
+from tango.common.file_lock import FileLock
+from tango.common.params import Params
+from tango.common.util import tango_cache_dir
+from tango.step import Step
+from tango.step_cache import CacheMetadata, StepCache
+from tango.step_caches.local_step_cache import LocalStepCache
+from tango.step_info import StepInfo
+
+from .common import Constants, step_dataset_name
+
+logger = logging.getLogger(__name__)
+
+
+@StepCache.register("beaker")
+class BeakerStepCache(LocalStepCache):
+    """
+    This is a :class:`~tango.step_cache.StepCache` that's used by :class:`BeakerWorkspace`.
+    It stores the results of steps on Beaker as datasets.
+
+    It also keeps a limited in-memory cache as well as a local backup on disk, so fetching a
+    step's resulting subsequent times should be fast.
+
+    :param workspace: The name or ID of the Beaker workspace to use.
+    :param beaker: The Beaker client to use.
+
+    .. tip::
+        Registered as :class:`~tango.step_cache.StepCache` under the name "beaker".
+    """
+
+    def __init__(self, workspace: Optional[str] = None, beaker: Optional[Beaker] = None):
+        super().__init__(tango_cache_dir() / "beaker_cache")
+        self.beaker: Beaker
+        if beaker is not None:
+            self.beaker = beaker
+            if workspace is not None:
+                self.beaker.config.default_workspace = workspace
+                self.beaker.workspace.ensure(workspace)
+        elif workspace is not None:
+            self.beaker = Beaker.from_env(default_workspace=workspace)
+        else:
+            self.beaker = Beaker.from_env()
+
+    def _acquire_step_lock_file(self, step: Union[Step, StepInfo], read_only_ok: bool = False):
+        return FileLock(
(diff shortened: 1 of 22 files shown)


tango/integrations/beaker/step_cache.py
```diff
@@ -0,0 +1,178 @@
...
+            # Make sure the step wasn't cached since the last time we checked (above).
+            if self.step_dir(step).is_dir():
+                return load_and_return()
+
+            dataset = self._step_result_dataset(step)
+            if dataset is None:
+                raise KeyError(step)
+
+            # We'll download the dataset to a temporary directory first, in case something goes wrong.
+            temp_dir = tempfile.mkdtemp(dir=self.dir, prefix=key)
```
dirkgr, line 133: This is fine, but `tempfile.TemporaryDirectory` exists and gives a higher-level API.

dirkgr: Ah, I guess `TemporaryDirectory` doesn't have `rmtree` capability.

tango/integrations/beaker/step_cache.py
```diff
@@ -0,0 +1,178 @@
...
+    def __len__(self) -> int:
+        return len(
+            [
+                ds
+                for ds in self.beaker.workspace.datasets(
+                    uncommitted=False, match=Constants.STEP_DATASET_PREFIX
+                )
+                if ds.name is not None and ds.name.startswith(Constants.STEP_DATASET_PREFIX)
+            ]
+        )
```
dirkgr, line 178:
```suggestion
        return sum(
                1
                for ds in self.beaker.workspace.datasets(
                    uncommitted=False, match=Constants.STEP_DATASET_PREFIX
                )
                if ds.name is not None and ds.name.startswith(Constants.STEP_DATASET_PREFIX)
        )
```

This saves some memory allocations because it never creates a temporary list.

tango/integrations/beaker/workspace.py
```diff
@@ -0,0 +1,278 @@
...
+        dataset_name = step_dataset_name(step_info)
+
+        step_info_dataset: Dataset
+        try:
+            step_info_dataset = self.beaker.dataset.create(dataset_name, commit=False)
+        except DatasetConflict:
+            step_info_dataset = self.beaker.dataset.get(dataset_name)
+
+        if step_info_dataset.committed:
+            return
```
dirkgr, line 271: What does this mean? If it's committed, we silently don't update the step info?

tango/step_info.py
```diff
@@ -157,7 +157,7 @@ def to_json_dict(self) -> Dict[str, Any]:
                 if isinstance(v, set)
                 else v
             )
-            for k, v in asdict(self).items()
+            for k, v in sorted(asdict(self).items(), key=lambda x: x[0])
```
dirkgr, line 160: Why sorted?

tango/integrations/beaker/workspace.py
```diff
@@ -0,0 +1,278 @@
...
+                f"Beaker dataset at {self.beaker.dataset.url(step_dataset_name(step))}"
+            )
+
+        # Update StepInfo to mark as running.
+        try:
+            step_info.start_time = utc_now_datetime()
+            step_info.end_time = None
+            step_info.error = None
+            step_info.result_location = None
+            self._update_step_info(step_info)
```
dirkgr, line 123: There is a race condition here. Do we have that race condition in the other workspaces as well?

dirkgr (changes requested):
This looks quite interesting! Will try it next time I run something.

tango/integrations/beaker/workspace.py
epwalsh: Yea.. I don't think we need that though. This shouldn't happen. 0eba6ab

tango/step_info.py
epwalsh: I was trying to make it deterministic so we get the same digest if it hasn't changed, but I just realized we'll need to return `OrderedDict` for that to work 🤦‍♂️ 0eba6ab

tango/integrations/beaker/workspace.py
epwalsh: You're right. I have an idea for a remote lock on Beaker. `LocalWorkspace` doesn't have this issue of course, but `WandbWorkspace` might.

epwalsh: There's another issue here: Workspaces in generally don't handle these race conditions well. They'll also result in a `RuntimeError` or some other (potentially silent) failure. This is my proposed fix for that: e1bf551

epwalsh: Remote lock: 1bc7cc3

epwalsh:
https://github.com/allenai/tango/pull/296/commits/49242aa13493988207eb69863e732d2c7021890c was a huge improvement, but https://github.com/allenai/cached_path/issues/95 would allow us to speed up the `registered_runs()` method even more.

epwalsh:
@dirkgr, you were right about only needing a remote lock, no local lock required. bdc969f

dirkgr (approved):

PR #304, opened 2022-06-01
Registerables first steps eg

### LINKED ISSUE #302: Adding a code example to the Registrable section in First Steps doc
**Is your feature request related to a problem? Please describe.**
When initially going through the [First Steps](https://ai2-tango.readthedocs.io/en/latest/first_steps.html) guide, I found the [Registrable section](https://ai2-tango.readthedocs.io/en/latest/first_steps.html#registrable) to be less clear than the rest of the document (though all of it is quite good!) I did find the in-text example of `TorchTrainerStep` taking registrable classes such as `Model` and `Optimizer` as inputs good for motivating the need to swap out subclasses frictionlessly. But without a code example to accompany it, it was still unclear to me what Registrables were doing to address this.

After using the library a bit, it was clear to me that without registrables researchers would have to change lots of type hints in code rather than just changing the config.jsonnet file to make experimental variations. But other than for the step objects (which are a bit more mysterious than just user defined objects), this dynamic does not appear in the code examples. Thus when I first read this, I was actually unaware that a config.jsonnet could have a field of a Step populated by another dictionary with its own "type" field (as in the example proposed below).

**Describe the solution you'd like**
I propose adding a code example to this section of the guide. It could be something along the lines of the following:

```json
{
  "steps": {
    "train": {
      "type": "train",
      "model": {
        "type": "architecture1", # or "architecture2",
        ...
      }
    }
  }
}
```

```python
from tango import Step
from tango.common import FromParams, Registerable

class Model(Registerable):
    ...

@Model.register("architecture1")
class Architecture1(Model):
    ...

@Model.register("architecture2")
class Architecture2(Model):
    ...

@Step.register("train")
class TrainerStep(Step):
    def run(self, model: Model) -> Model:
        ...
```

**Describe alternatives you've considered**
I'd also considered making an abstract foo, bar, baz style example. This has the advantage that it is possible to include complete code without ellipses. However I do really like the existing motivation provided by using the case of a trainer and its inputs.

jagnusson:
Closes #302 

- Adds a config and code example of registrable to the First Step docs
- Rewrites the Registrable section running text to emphasize how this build on the motivation for FromParams additionally allowing subclasses to be specified within nested specifications of a config file.

## Before submitting

## After submitting

### DIFF
--- a/docs/source/first_steps.md
+++ b/docs/source/first_steps.md
@@ -372,9 +372,9 @@ This is also very useful when you're writing a step that requires a certain type
 
 ```python
 from tango import Step
-from tango.common import FromParams, Registerable
+from tango.common import FromParams, Registrable
 
-class Model(torch.nn.Module, Registerable):
+class Model(torch.nn.Module, Registrable):
     ...
 
 @Model.register("variant1")
--- a/docs/source/first_steps.md
+++ b/docs/source/first_steps.md
@@ -362,25 +362,6 @@ class FooStep(Step):
         return sum([bar.x for bar in bars]) + baz.bar.x
 ```
 
-And the corresponding config could look like this:
-
-```json
-{
-  "steps": {
-    "foo": {
-      "type": "foo",
-      "bars": [
-        {"x":1},
-        {"x":2}
-      ],
-      "baz": {
-        "bar": {"x": 3}
-      }
-    }
-  }
-}
-```
-
 ### `Registrable`
 
 The {class}`~tango.common.registrable.Registrable` class is a special kind of {class}`~tango.common.from_params.FromParams` class that allows you specify from the config which subclass of an expected class to deserialize into.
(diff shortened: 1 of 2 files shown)


docs/source/first_steps.md
```diff
@@ -364,16 +364,48 @@ class FooStep(Step):
 
 ### `Registrable`
 
-The {class}`~tango.common.registrable.Registrable` class is special kind of {class}`~tango.common.from_params.FromParams` class that allows you to deserialize
-any subclass of the expected class from a config.
+The {class}`~tango.common.registrable.Registrable` class is a special kind of {class}`~tango.common.from_params.FromParams` class that allows you specify from the config which subclass of an expected class to deserialize into.
```
epwalsh, line 367:
```suggestion
The {class}`~tango.common.registrable.Registrable` class is a special kind of {class}`~tango.common.from_params.FromParams` class that allows you to specify from the config which subclass of an expected class to deserialize into.
```

PR #307, opened 2022-06-04
Make tests runnable out-of-tree for help with conda-packaging

h-vetinari:
This is the equivalent of https://github.com/allenai/allennlp/pull/5560, but for tango.

I've been carrying a simpler version of this patch since the beginning of the feedstock (since conda does not run the tests in a source checkout, but for the finished package) - this version allows introduces an override also for other users along the model introduced for allennlp.

However, as of 0.7.0 this is not complete (and I've been struggling with this in https://github.com/conda-forge/tango-feedstock/pull/8), because `multicore_executor_test` does not respect this infrastructure. I'm going to need some help on the latter (unless I just unconditionally skip all tests from that module) - my first impression is that `test_fixtures.package.steps` should be in some auxiliary module as part of tango, _not_ in the fixtures.


epwalsh:
I'm certainly okay with moving the test steps into the tango module somewhere. Maybe at `tango.common.testing` to be consistent with allennlp. But we should make sure to exclude them from being built into the package (just need to add a line to the `setup.py`).

h-vetinari:
> I'm certainly okay with moving the test steps into the tango module somewhere. Maybe at `tango.common.testing` to be consistent with allennlp.

Cool!

> But we should make sure to exclude them from being built into the package (just need to add a line to the `setup.py`).

Then I still don't see a way how the multicore tests should be runnable out-of-tree...

epwalsh:
I guess I misunderstood how the tests need to run in this case. But that's fine, it's okay to include those files in the build if we have to.

h-vetinari:
@epwalsh @dirkgr 
I'm going to need help on this one, especially the multicore tests (or an instruction from you to skip them all).

dirkgr:
What's the problem here? It looks like the tests failed that need secrets, probably because @h-vetinari is not a member of the AI2 org. That's not a real failure though. Is there anything wrong with the multicore tests now?

h-vetinari:
> What's the problem here?

The issue is that the multicore tests cannot run out of tree due to the way they import the `Step`s. From the OP:

> However, as of 0.7.0 this is not complete (and I've been struggling with this in [conda-forge/tango-feedstock#8](https://github.com/conda-forge/tango-feedstock/pull/8)), because `multicore_executor_test` does not respect this infrastructure. I'm going to need some help on the latter (unless I just unconditionally skip all tests from that module) - my first impression is that `test_fixtures.package.steps` should be in some auxiliary module as part of tango, _not_ in the fixtures.

epwalsh:
Hey @h-vetinari, per the above conversation I thought we had decided to move the test steps into a proper tango submodule, like `tango.common.testing`. Does that sound good? If so, do you want to make that change in this PR?

h-vetinari:
> Does that sound good?

Yes!

> If so, do you want to make that change in this PR?

Sorry, been pretty busy and only getting to this now.

h-vetinari:
The 0.7.0 version of these patches now passed the test suite in https://github.com/conda-forge/tango-feedstock/pull/8. :)

If/once you approve this, I'll be happy to rebuild the other versions in between.

h-vetinari:
PS. I needed to skip `TestRun.test_deterministic_experiment` though - the test hardcodes specific hashes for the workers(?) and this is simply not reproducible with a slightly different setup, I believe.

h-vetinari:
Ping @dirkgr @epwalsh

tango/common/testing.py
```diff
@@ -145,3 +172,146 @@ def requires_gpus(test_method):
             test_method
         )
     )
+
```
epwalsh, line 175: Can you move these to a separate module? I don't want these steps to necessarily be registered when someone just wants to import the `TangoTestCase` class.

I would suggest making `tango/common/testing/` directory, and move this file to `tango/common/testing/__init__.py`, then have a separate file `tango/common/testing/steps.py` where you put these steps (don't import anything from `steps.py` into `__init__.py`).

epwalsh (changes requested):
Hey @h-vetinari, sorry for the slow response (Dirk is on vacation this week). I just have an organization suggestion.

tango/common/testing.py
h-vetinari: Thanks for the review - no worries, I was slow too!

> I don't want these steps to necessarily be registered when someone just wants to import the `TangoTestCase` class.

I don't see how that would happen except with `*` imports, which are discouraged anyway?

I mean I can do the change of course, I just don't really understand the gain...

epwalsh: If those test steps are in the same file as `TangoTestCase` they will be registered when you import `TangoTestCase` even if you don't import those steps since every global statement in the file, including the `@Step.register()` statements, will be evaluated on import.

You can check this:

```python
from tango.step import Step

assert Step.list_available() == []

from tango.common.testing import TangoTestCase

assert "float" in Step.list_available()
```

h-vetinari: Actually, there is one substantial advantage of this approach - the diff gets much smaller. ;-)

h-vetinari: > If those test steps are in the same file as `TangoTestCase` they will be registered when you import `TangoTestCase` even if you don't import those steps since every global statement in the file, including the `@Step.register()` statements, will be evaluated on import.

Ah, I was not aware there were some global shenanigans going on 😅 

Also, it seems our responses were seconds apart. ;-)

epwalsh:
Looks like there's some more imports that need to be updated in tests.

h-vetinari:
> Looks like there's some more imports that need to be updated in tests.

Feel free to push into this PR. I'm not at a computer right now, and waiting for approval for CI each time drags out finding these trivial issues quite a bit 😅
(No idea why it didn't fail in the previous commit, I thought I had fixed everything)

epwalsh:
Cool I'll take a look. Can you make sure you have the "Allow edits from maintainers" option checked?

epwalsh (approved):
LGTM! There are still some failing tests but those are due to missing secrets because of GitHub security. Once this merges into main those tests should pass. Let me know if you're satisfied with everything and I'll hit the merge button!

h-vetinari:
> Let me know if you're satisfied with everything and I'll hit the merge button!

Yeah this looks good, thanks a lot for pushing through the final clean ups!

PR #312, opened 2022-06-09
Fix bug in Beaker workspace

epwalsh:
Fixes https://github.com/allenai/catwalk/issues/35

Changes proposed:
- `BeakerWorkspace.step_info(step)` will now create the step info for `step` if it doesn't exist yet, consistent with how our other workspaces work.


epwalsh:
> Ah, so this happened because I wasn't using a config file?

Yes, since using from a config file would result in the step info being created when `register_run()` is called.

PR #314, opened 2022-06-13
fix non-deterministic behavior in TorchTrainStep

lgatys:
This PR makes the TorchTrainStep deterministic.

- Moves the `set_seed_all` call to the beginning of the `_train` function to run before model initialization. 

## Before submitting

## After submitting

### DIFF
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -348,6 +348,14 @@ def _train(
     callbacks: Optional[List[Lazy[TrainCallback]]] = None,
     include_package: Optional[Set[str]] = None,
 ) -> Optional[Model]:
+    # Set random seeds.
+    set_seed_all(config.seed)
+
+    # force torch to be deterministic
+    os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
+    torch.backends.cudnn.benchmark = False
+    torch.use_deterministic_algorithms(True)
+
     config.worker_id = worker_id
 
     if config.is_distributed and include_package:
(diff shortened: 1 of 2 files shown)


tango/integrations/torch/train.py
```diff
@@ -348,6 +348,14 @@ def _train(
     callbacks: Optional[List[Lazy[TrainCallback]]] = None,
     include_package: Optional[Set[str]] = None,
 ) -> Optional[Model]:
+    # Set random seeds.
+    set_seed_all(config.seed)
+
+    # force torch to be deterministic
+    os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
+    torch.backends.cudnn.benchmark = False
+    torch.use_deterministic_algorithms(True)
```
epwalsh, line 357: This has a negative effect on performance, so we probably don't want to do this. If you need this you could always call it from a callback though.

epwalsh (changes requested):
Looks good! Just one comment.

tango/integrations/torch/train.py
lgatys: Makes sense, I like the callback idea! 
But should the step then still be flagged as `DETERMINISTIC` ? It could be confusing because the training results can vary substantially without these calls.

epwalsh: In my opinion, yes, because there are benefits of having the step be marked DETERMINISTIC: caching isn't useful for non DETERMINISTIC steps. Besides, even after calling this torch function you're not guaranteed deterministic results.

epwalsh (approved):
Thanks Leon!

lgatys:
> Thanks Leon!

No worries! Just FYI: I won't be able to merge PR because of restrictions.

PR #315, opened 2022-06-15
Beaker DOS fix

dirkgr:


epwalsh (approved):

tango/integrations/beaker/workspace.py
```diff
@@ -292,7 +295,10 @@ def _get_run_from_dataset(self, dataset: Dataset) -> Optional[Run]:
 
         import concurrent.futures
 
-        with concurrent.futures.ThreadPoolExecutor() as executor:
+        with concurrent.futures.ThreadPoolExecutor(
+            max_workers=50,
```
epwalsh, line 299: Actually this probably makes the situation worse. The default is not more than 36.
https://github.com/python/cpython/blob/e929dae42ad6aaf55e5cce6339c1ef97c01c8f7b/Lib/concurrent/futures/thread.py#L142

epwalsh (changes requested):

tango/integrations/beaker/workspace.py
dirkgr: Oh. Then the documentation for that is wrong.

epwalsh: Yeup 🙃

dirkgr:
@epwalsh good to go?

epwalsh:
Can we make max workers a configurable parameter so this isn't always the default? Or we just need to remember to roll this back once it no longer breaks Beaker.

dirkgr:
Since the limit is pretty much a function of the beaker service, I think it should be hardcoded, just like other aspects of the beaker service are hardcoded.

epwalsh:
Fair enough. Can we at least make an issue to roll it back once fixed so we don't forget?

epwalsh (approved):

dirkgr:
Tracking in https://github.com/allenai/tango/issues/317.

PR #316, opened 2022-06-15
Log from `BeakerStepLock` at WARNING level

epwalsh:


dirkgr (approved):
This is good, but tests don't pass.

dirkgr:
Should this be printed to the click logger? Do we still have that?

epwalsh:
> Should this be printed to the click logger? Do we still have that?

We still have that (it's the `cli_logger` now), but the downside would be that those messages wouldn't show up when using Tango programmatically unless the our `initialize_logging()` function was called with `enable_cli_logs=True`.

dirkgr:
`WARNING` is our cheap way of getting both at the same time? I think it's OK in this case.

epwalsh:
> WARNING is our cheap way of getting both at the same time? I think it's OK in this case.

Right

epwalsh:
> This is good, but tests don't pass.

Looks like Beaker is still broken 🙂

PR #323, opened 2022-06-24
Modernize our issue templates

epwalsh:
Following upstream changes from https://github.com/allenai/python-package-template/pull/57.

This change was prompted by @aimichal pointing out that our bug report issue template is way too complex, which adds friction for people to report bugs. This is what the new version looks like:

image


aimichal (approved):
Nice and friendly now!

dirkgr (approved):
I still find these fairly complex. I'm not sure we have the volume of engagement to go deep on these templates. But they are better than the previous version.

jagnusson (approved):
Looks great! Would it make sense to still point to CONTRIBUTING.md somewhere but frame it as extra guidance for those who want it rather than required steps for anyone to submit an issue?

isunitha98selvan:
Looks great. Might be helpful to hint somewhere to add in the expected behavior in the bug report.

epwalsh:
> Would it make sense to still point to CONTRIBUTING.md somewhere but frame it as extra guidance for those who want it rather than required steps for anyone to submit an issue?

@jagnusson that's a good point, but I think GitHub automatically notifies first-time contributors with a banner pointing to the CONTRIBUTING.md file when they open a PR. I noticed the other day that it actually notified me when I opened a PR that the CONTRIBUTING.md had changed since my last PR 🤯

isunitha98selvan (approved):

epwalsh:
@isunitha98selvan done: 1094c67

jagnusson:
> > Would it make sense to still point to CONTRIBUTING.md somewhere but frame it as extra guidance for those who want it rather than required steps for anyone to submit an issue?
> 
> @jagnusson that's a good point, but I think GitHub automatically notifies first-time contributors with a banner pointing to the CONTRIBUTING.md file when they open an issue or PR. I noticed the other day that it actually notified me when I opened a PR that the CONTRIBUTING.md had changed since my last PR 🤯

Great point! Yes that more than handles my concern.

PR #331, opened 2022-06-28
Metadata changes, make executor registrable

epwalsh:
This PRdoes some preliminary work before I introduce the `BeakerExecutor`.

- Removes `StepExecutionMetadata`. All of its fields are absorbed into `StepInfo`.

  Why? This dataclass didn't really fit into our API. It was defined in `workspace.py`, but wasn't actually part of the `Workspace` API. I think its odd standing was due to several refactors involving the `Executor` and `Workspace` abstractions. So, since this class already had some overlap with `StepInfo` and `StepInfo` *is* actually involved with the `Workspace` API, I decided to merge the data from this class into `StepInfo`.
- Makes `Executor` a registrable class that can be configured in a `tango.yml` settings file.

## Before submitting

## After submitting


tango/__main__.py
```diff
@@ -642,7 +641,7 @@ def _run(
     include_package: Optional[Sequence[str]] = None,
     start_server: bool = True,
     step_name: Optional[str] = None,
-    parallelism: int = 1,
+    parallelism: Optional[int] = None,
```
dirkgr, line 644: What does it mean if it's `None`? Should there be a doc change to go with it?

tango/common/util.py
```diff
@@ -264,3 +266,37 @@ def utc_now_datetime() -> datetime:
...
+        return {key: replace_steps_with_unique_id(value) for key, value in o.items()}
+    else:
+        return o
+
+
+def jsonify(o: Any) -> Any:
+    """
+    Transform an arbitrary object into a JSON-serializable equivalent (if there is one)
+    in a deterministic way.
+    """
```
dirkgr, line 288: I suspect you copied this from somewhere, so that's fine. But with a doc string like that I would also expect that it transforms single-item float tensors into `float`, because that's always an annoying problem too.

tango/step_info.py
```diff
@@ -1,13 +1,73 @@
...
+            if result.returncode != 0:
+                if result.stderr is not None:
+                    logger.exception("Error saving conda packages: %s", result.stderr.decode())
+                else:
+                    result.check_returncode()
+            elif result.stdout is not None:
+                return yaml.safe_load(result.stdout.decode())
+        except Exception as exc:
+            logger.exception("Error saving conda packages: %s", exc)
+    return None
```
dirkgr, line 70: What do you think about getting rid of this entire thing? I know in principle it would be quite useful to have for debugging. But in practice we almost never look at it, and it takes several seconds per step to do this. When the steps are fast, this is where 90% of the execution time is.

dirkgr (approved):
Only small comments. Looks good!

tango/step_info.py
epwalsh: That's fine with me. It seems too brittle anyway.

PR #340, opened 2022-07-19
Add `BeakerExecutor`

### LINKED ISSUE #123: Run in beaker
Goal: Run either some steps or maybe all steps in Beaker. Steps need a way to specify requirements for the number of CPUs, number of GPUs, amount of memory. 

Not a goal: Run multi-node jobs

This is not trivial, not least because we have to find a way to package up a project and stick it into a Docker image to run.

epwalsh:
Closes #123 by implementing the `BeakerExecutor`.

### External changes

See the CHANGELOG for a complete description of all user-facing changes.

### Internal changes

- Many logging tweaks and improvements, including:
  - Unifying how we log the state of a step throughout its execution with the `cli_logger`.
  - Improving how we log exceptions. Executors and workspaces don't need to worry about this.
- Added new method to `Executor` class: `execute_sub_graph_for_step()`, which is now called from `__main__` when executing a single step instead of `execute_step()`. I did this so that the `BeakerExecutor` would always get the `StepGraph` object even when running a single step, which it uses to serialize the step and its dependencies (I also experimented with using dill to steps, but that has some big downsides).

## Implementation notes

The toughest part of implementing the Beaker executor was seamlessly streaming logs from a Beaker job for a step to a user's local terminal in a way that preserves all of the metadata in each log record, including exceptions and tracebacks. So I decided to implement a logging `Handler` that serializes `LogRecord` objects as JSON. Steps that run in Beaker use this handler, and then the executor deserializes those records locally and emits them with whatever handlers are configured locally. The downside is that logs are not really human-readable on the Beaker dashboard.

## Before submitting

## After submitting


CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+      beaker_workspace: ai2/my-workspace
+      clusters:
+        - ai2/general-cirrascale
+    ```
+    See the docs for the `BeakerExecutor` for more information on the input parameters.
+
+### Changed
+
+- **CLI:**
+  - The `tango run` command will throw an error if you have uncommitted changes in your repository, unless
```
dirkgr, line 35: Can we do this only if you use the Beaker executor? I think that's a pretty big obstacle for rapid development for many people. Constraining people's workflow always makes the library feel heavy.

epwalsh: > Can we do this only if you use the Beaker executor?

I guess, but I don't think this will constrain anyone's workflow. Let's say you have some uncommitted changes and go:

```bash
> tango run ...
ERROR: You have uncommitted changes! Use --allow-dirty to force.
```

At first you're like "what the heck, that's annoying!" But it makes you pause and think "should I actually commit these changes?" If the answer to that is still "no", then you just hit the up-arrow key and append `--allow-dirty` to the last command. That's it, all good. Next time you hit the up-arrow for `tango run ...` again it will already have the `--allow-dirty` flag. So, at worst, it's an extremely minor inconvenience that costs you 3-5 seconds depending on how quickly you type, at best it saves you from not knowing what state your code was actually in when you ran that experiment.

Also: we save the `git` state of the repository in each `StepInfo` object (an important part of the user-facing API of Tango), so I think it's important that we nudge users to have a clean, committed working directory when they run steps, otherwise the `git` state in their `StepInfo` objects will be incorrect.

tango/__main__.py
```diff
@@ -267,12 +267,10 @@ def cleanup(*args, **kwargs):
...
     type=int,
-    help="""The maximum number of steps to run in parallel (if possible).
-    By default there is no parallelism, but each step is ran in its own subprocess (same as setting '-j=1').
-    A value of 0 or less means each step is ran in the main process using the default executor.
-    A value greater than 1 means at most 'j' steps will be ran in parallel,
-    each within its own subprocess, using the multicore executor.
-    This option is ignored if you've configured your executor in a 'tango.yml' settings file.""",
+    help="""The maximum number of steps to run in parallel (for executors that support this).
+    The exact behavior depends on the executor. If you're using the default executors,
+    a value of 0 (or left unspecified) means each step is ran in the main process using the default executor,
```
AkshitaB, line 272:
```suggestion
    a value of 0 (or left unspecified) means each step is run in the main process using the default executor,
```

tango/__main__.py
```diff
@@ -287,6 +285,11 @@ def cleanup(*args, **kwargs):
     type=str,
     help="""Specify the name for this run.""",
 )
+@click.option(
+    "--allow-dirty",
+    is_flag=True,
+    help="""Allow running the experiment with a dirty working directory.""",
```
AkshitaB, line 291:
```suggestion
    help="""Allow running the experiment with a dirty working directory (uncommitted code changes).""",
```

tango/__main__.py
```diff
@@ -331,9 +335,62 @@ def run(
...
+        for package_name in include_package:
+            import_extra_module(package_name)
+
+    # Load step graph and step.
+    step_graph = StepGraph.from_file(experiment)
+    step = step_graph[step_name]
+
+    # Initialize workspace and executor.
+    workspace = Workspace.from_url(workspace_url)
+    executor = Executor(workspace=workspace, include_package=include_package)
```
AkshitaB, line 384: So the `BeakerExecutor` uses the default executor when running individual steps?

tango/common/exceptions.py
```diff
@@ -35,12 +35,26 @@ class RegistryKeyError(ConfigurationError):
     """
 
 
-class SigTermReceived(TangoError):
+class CancellationError(TangoError):
+    """
+    Base class for errors raised due to manually cancellation of a run or step.
```
AkshitaB, line 40:
```suggestion
    Base class for errors raised due to manual cancellation of a run or step.
```

tango/executor.py
```diff
@@ -46,9 +51,11 @@ def __init__(
         self,
         workspace: Workspace,
         include_package: Optional[Sequence[str]] = None,
+        parallelism: Optional[int] = None,
     ) -> None:
         self.workspace = workspace
         self.include_package = include_package
+        self.parallelism = parallelism
```
AkshitaB, line 58: We should add a log statement here if `parallelism` is anything other than `None`, essentially saying what the note in the docstring says.

tango/executor.py
```diff
@@ -88,14 +93,23 @@ def execute_step_graph(
...
+                    import rich
 
-        for stacktrace in error_tracebacks:
-            logger.error(stacktrace)
+                    failed.add(step.name)
+                    logger.exception(exc, extra={"highlighter": rich.highlighter.ReprHighlighter()})
 
         return ExecutorOutput(successful=successful, failed=failed, not_run=not_run)
 
+    def execute_sub_graph_for_step(
```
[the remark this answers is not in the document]
AkshitaB: I'm curious, why is the step_graph easier to serialize? Wouldn't serializing the step graph also require serializing the individual steps too?

tango/executor.py
```diff
@@ -88,14 +94,21 @@ def execute_step_graph(
...
+        self, step_graph: StepGraph, step_name: str, run_name: Optional[str] = None
+    ) -> None:
+        """
+        Execute the sub-graph associated with a particular step in a
+        :class:`~tango.step_graph.StepGraph`.
+        """
+        step = step_graph[step_name]
```
AkshitaB, line 110: We should add more info in the docstring here explaining why we need this separate method -- it'll be easy to lose/forget the information discussed in this PR.

tango/executors/multicore_executor.py
```diff
@@ -217,7 +218,7 @@ def _try_to_execute_next_step(config_path: str, run_name: Optional[str] = None)
             if len(_queued_steps) == 0:
                 logger.debug("No steps in queue!")
                 return
-            if len(_running) < self.parallelism:
+            if len(_running) < (self.parallelism or 1):
```
AkshitaB, line 221: Why the extra check with `or 1` if we've already updated `self.parallelism` on line 36?

tango/executors/multicore_executor.py
```diff
@@ -239,7 +240,7 @@ def _try_to_execute_next_step(config_path: str, run_name: Optional[str] = None)
                 _running[step_name] = process
             else:
                 logger.debug(
-                    f"{self.parallelism} steps are already running. Will attempt to execute later."
+                    f"{self.parallelism or 1} steps are already running. Will attempt to execute later."
```
AkshitaB, line 243: Same question as before.

tango/integrations/beaker/executor.py
```diff
@@ -0,0 +1,575 @@
...
+        # Keeps track of steps that are next up to run on Beaker.
+        steps_to_run: Set[str] = set()
+        # These are steps that have been submitted to Beaker but haven't completed yet.
+        submitted_steps: Set[str] = set()
+        # Futures for tracking the Beaker runs for each step.
+        step_futures: List[concurrent.futures.Future] = []
+
+        uncacheable_leaf_steps = step_graph.uncacheable_leaf_steps()
+
+        def update_steps_to_run():
```
AkshitaB, line 209: Maybe we should shift some of this logic to some `utils`. Just so our logic for updating steps remains the same across `MulticoreExecutor` and `BeakerExecutor`.

tango/integrations/beaker/executor.py
```diff
@@ -0,0 +1,575 @@
...
+    def _check_if_cancelled(self):
+        if self._is_cancelled.is_set():
+            raise RunCancelled
+
+    def _execute_sub_graph_for_step(
+        self,
+        step_graph: StepGraph,
+        step_name: str,
+        run_name: Optional[str],
+        in_thread: bool = False,
```
AkshitaB, line 308: In essence, `in_thread` is analogous to `called_by_executor` in the `__main__` script, correct? Maybe at some point, I should refactor `multicore` to handle it cleanly like this too.

tango/step.py
```diff
@@ -58,6 +60,40 @@ def get_args(tp):  # type: ignore
...
+    """
+    TaskResources describe minimum external hardware requirements which must be available for a
+    step to run.
+    """
+
+    cpu_count: Optional[float] = None
+    """
+    Minimum number of logical CPU cores. It may be fractional.
+
+    Examples: ``4``, ``0.5``.
```
AkshitaB, line 74: What do we mean by fractional cpu cores?

tango/__main__.py
epwalsh: Ugh lol this is 2nd time you've caught me with this typo 🤦‍♂️

tango/__main__.py
epwalsh: Yes, that's worth a comment. I'll add one.

tango/executor.py
epwalsh: Hmm that's a good point, but if we logged from `__init__()` we'd see that log statement from every subclass as well since the subclasses call `super().__init__()`. But I could put a warning/log statement in `execute_step_graph()` instead.

tango/executor.py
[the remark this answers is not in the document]
epwalsh: I should clarify: we serialize the step config, accessed via the step graph.

tango/executors/multicore_executor.py
epwalsh: Just for safety since `self.parallelism` is technically a public attribute, and so can be easily modified in-place.

tango/step.py
epwalsh: Beaker allows this, and I think it's just passed on to the `docker run ...` command as-is.

https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler

tango/integrations/beaker/executor.py
epwalsh: Yeup

AkshitaB (commented):
Looks good to me, overall. I think it'll be helpful in the docs to have an example for using the beaker config on-the-fly too. Eg.

`tango --settings beaker.yml run config -w <workspace>` etc.

Also, I noticed something interesting.

I run `tango --settings beaker.yml run test_fixtures/integrations/torch/train.jsonnet -w localworkspace -i test_fixtures/integrations/torch/__init__.py -i test_fixtures/__init__.py`

1. I don't have the cuda version of torch installed locally, so the run gave me an error, as by default, it was expecting to run on a `gpu` but the torch version it installed on beaker wasn't gpu-compatible. I had to specify a gpu count of 0 to get it to work. Is there a way to handle this?
2. I specified `-w localworkspace` out of habit, and it ran, but didn't cache results. So, `localworkspace` cache was empty, and I had no way of getting the output anywhere. This is unexpected behavior, so some more clarity around this in docs/logs will be helpful.

```[step train] Loading best weights from                                                                                                              train.py:336
                             /Users/akshitab/local/code/tango/localworkspace/cache/TorchTrainStep-i453ThuNLoK2Z65P8emxmEfZxYkYgCrm/work/weights.pt                                           
[08/01/22 17:26:15] INFO     ✓ Finished run new-ape                                                                                                                           __main__.py:841
                                                                                                                                                                                             
 ┏━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┓                                                                                                                                                 
 ┃ Step Name ┃ Status      ┃ Cached Result ┃                                                                                                                                                 
 ┡━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━┩                                                                                                                                                 
 │ train     │ ✓ succeeded │ N/A           │                                                                                                                                                 
 └───────────┴─────────────┴───────────────┘                                                                                                                                                 
                ✓ 1 succeeded                                                                                                                                                                
                                            ```

PR #344, opened 2022-07-26
Fix issue where StepInfo config could be parsed into a Step

lgatys:

- Prevents `construct_arg` method in `from_params.py` to try to parse the argument as a Step, if it is the `config` argument of the `StepInfo` class. 
- This happened when the `config` contained references to upstream steps, which evaluated the `could_be_step` condition to `True`.
- The previous behavior was creating issues with the WandbWorkspace when trying to update and re-parse the StepInfo.

## Before submitting

## After submitting

### DIFF
--- a/tests/step_info_test.py
+++ b/tests/step_info_test.py
@@ -31,12 +31,12 @@ def test_step_info():
 def test_step_info_with_step_dependency():
     """Checks that the StepInfo config is not parsed to a Step if it has dependencies on upstream steps"""
 
-    @Step.register("foo")
+    @Step.register("foo", exist_ok=True)
     class FooStep(Step):
         def run(self, bar: Any) -> str:  # type: ignore
             return "foo" + bar
 
-    @Step.register("bar")
+    @Step.register("bar", exist_ok=True)
     class BarStep(Step):
         def run(self) -> str:  # type: ignore
             return "Hey!"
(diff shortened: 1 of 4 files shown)


tests/step_info_test.py
```diff
@@ -23,3 +26,35 @@ def test_step_info():
...
     deserialized = StepInfo.from_json_dict(json.loads(serialized))
     assert deserialized == step_info
+
+
+def test_step_info_with_step_dependency():
+    """Checks that the StepInfo config is not parsed to a Step if it has dependencies on upstream steps"""
+
+    @Step.register("foo")
+    class FooStep(Step):
+        def run(self, bar: Any) -> str:
```
epwalsh, line 36:
```suggestion
        def run(self, bar: Any) -> str:  # type: ignore
```

tests/step_info_test.py
```diff
@@ -23,3 +26,35 @@ def test_step_info():
...
+            return "foo" + bar
+
+    @Step.register("bar")
+    class BarStep(Step):
+        def run(self) -> str:
```
epwalsh, line 41:
```suggestion
        def run(self) -> str:  # type: ignore
```

epwalsh (commented):
Just a couple minor comments!

tests/step_info_test.py
epwalsh: This will tell `mypy` to stop complaining the "Lint" checks.

tests/step_info_test.py
```diff
@@ -23,3 +26,35 @@ def test_step_info():
...
+            return "Hey!"
```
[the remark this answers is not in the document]
lgatys: Done!

epwalsh (approved):
LGTM! I'll make a patch release once this merges

PR #348, opened 2022-07-27
Put PIP install instructions first

epwalsh:
Our conda installation method seems to break easily and is not always easy to debug or fix since we don't have direct control over it, so PIP should be the recommended way to install Tango.

- Switches the order of the install instructions. PIP is listed first now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,26 @@ AI2 Tango replaces messy directories and spreadsheets full of file versions by o
 
 **ai2-tango** requires Python 3.7 or later.
 
+### Installing with `pip`
+
+**ai2-tango** is available [on PyPI](https://pypi.org/project/ai2-tango/). Just run
+
+```bash
+pip install ai2-tango
+```
+
+To install with a specific integration, such as `torch` for example, run
+
+```bash
+pip install 'ai2-tango[torch]'
+```
+
+To install with all integrations, run
+
+```bash
+pip install 'ai2-tango[all]'
+```
+
 ### Installing with `conda`
 
 **ai2-tango** is available on conda-forge. You can install just the base package with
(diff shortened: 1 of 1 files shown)


schmmd (approved):
Thanks for changing this.  This was actually on mind randomly during my backpacking trip last week.  I'm fine with just the reordering, but I think we should remove the conda instructions as they are not working presently.  I have no stake in which way is better though--and the most important thing is simply that our recommended approach is first!

epwalsh:
That's fine with me but I think Dirk should weigh in since he's a fan of the conda install

h-vetinari:
> [...] is not always easy to debug or fix since we don't have direct control over it, so PIP should be the recommended way to install Tango.

Debugging is only a pull-request to https://github.com/conda-forge/tango-feedstock away. ;-)
Also, I'd claim that I've made every effort to make the collaboration work as smoothly as possible for you (and extended the invitation to join the maintenance of the feedstock), so I do think the "no direct control" is an unfair characterisation TBH.

I will happily defend the point that conda is technically superior to pip, but it's true that there can be delays through the necessary integration work, in particular if I don't get prompt help with stuff like https://github.com/allenai/tango/pull/307, or when you add new dependencies that aren't in conda-forge yet (and don't let me know ahead of time).

In any case, tango builds are now again up-to-date in conda-forge.

epwalsh:
@h-vetinari, you've done an excellent job and we really appreciate everything!

And I agree, conda is a better package manager than pip. But having to maintain a separate repo just to get things deployed on conda is not ideal. The other problem is that most of us don't know much about maintaining a conda package, so we always have to defer to you on issues instead of actively fixing things ourselves. I'll try to find some time to learn more about the process so I can be more help.

PR #361, opened 2022-08-15
Fix #355 - ensure git metadata is up-to-date

### LINKED ISSUE #355: Need to delete workspace when changing requirements
### 🐛 Describe the bug

When using the Beaker executor, I need to use a fresh workspace whenever I change requirements or it will used the requirements I had specified in the previous run.  This might be by design...

### Versions

Latest

epwalsh:
Fixes #355

### DIFF
--- a/tango/integrations/beaker/executor.py
+++ b/tango/integrations/beaker/executor.py
@@ -33,6 +33,7 @@ from tango.common.logging import cli_logger, log_exception, log_record_from_json
 from tango.executor import Executor, ExecutorOutput
 from tango.step import Step
 from tango.step_graph import StepGraph
+from tango.step_info import GitMetadata
 from tango.version import VERSION
 from tango.workspace import Workspace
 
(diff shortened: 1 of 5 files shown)


tango/workspaces/local_workspace.py
```diff
@@ -251,6 +251,10 @@ def step_starting(self, step: Step) -> None:
                 f"file at {self._step_lock_file(step)}.",
             )
 
+        if step_info.state == StepState.FAILED:
+            # Refresh environment metadata since it might be out-of-date now.
+            step_info.refresh()
```
dirkgr, line 256: Why only when it failed?

dirkgr (approved):

tango/integrations/beaker/executor.py
```diff
@@ -597,21 +598,22 @@ def _build_experiment_spec(self, step_graph: StepGraph, step_name: str) -> Exper
         )
 
         # Ensure we're working in a GitHub repository.
+        git = GitMetadata.check_for_repo()
```
schmmd, line 601: This change makes a lot of sense to me!

tango/workspaces/local_workspace.py
epwalsh: I think this is the only situation where it could be out-of-sync. @schmmd discovered this edge case. It happens like this:

1. Run a step that fails or cancel it as it's running.
2. Change something in your code or requirements without changing the VERSION of the step (perfectly reasonable thing to do).
3. Run it again.

If the step succeeds the environment metadata in step info is now out-of-date.

PR #362, opened 2022-08-16
Add `Step.RESULT_KIND` class variable

epwalsh:
Workspaces can use this field to categorize step results. The `WandbWorkspace` uses this to set the artifact kind field.

This was motivated by a discussion I had today with @cvphelps and others. For now only the W&B workspace uses this field. It allows Tango+W&B users to organize their step results in the dashboard better. And by setting the artifact kind for model outputs (like from the `TorchTrainStep`) to "model", those artifacts will automatically be added to W&B's model zoo (which is internal to our org). We might to come up with other standard result kinds besides "model", like for dataset creation/preprocessing steps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
         - ai2/general-cirrascale
     ```
     See the docs for the `BeakerExecutor` for more information on the input parameters.
+- **Step class**:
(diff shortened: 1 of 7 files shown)


dirkgr (changes requested):
I'm hesitant to put something into the step class that's so specific to WandB. The Tango API is already getting very big.

Can we just add a metadata field instead, `Dict[str, Any]`, where we keep special information for the WandB workspace? We could put other stuff there too, like the kinds of machine we want to run on for the Beaker executor.

epwalsh:
Closing in favor of https://github.com/allenai/tango/pull/363

PR #363, opened 2022-08-18
Add `workspace_metadata` arg to `Step` class, allow changing artifact kind in W&B workspace

epwalsh:
This allows users to set the artifact kind of step result artifacts saved to Weights & Biases (using the `WandbWorkspace`) by adding a field called `artifact_kind` to the `workspace_metadata` of a step.

### DIFF
--- a/tango/step.py
+++ b/tango/step.py
@@ -117,6 +117,8 @@ class Step(Registrable, Generic[T]):
       of inputs.
     :param step_resources: gives you a way to set the minimum compute resources required
       to run this step. Certain executors require this information.
+    :param workspace_metadata: use this to specify additional metadata for your workspace.
+      The type of metadata depends on your workspace.
 
     .. important::
         Overriding the unique id means that the step will always map to this value, regardless of the inputs,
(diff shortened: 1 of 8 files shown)


tango/step.py
```diff
@@ -242,6 +245,7 @@ def __init__(
         ] = None  # This is set only while the run() method runs.
         self._config = step_config
         self.step_resources = step_resources
+        self.workspace_metadata = workspace_metadata
```
dirkgr, line 248: If you add a member named `workspace_metadata` to the class directly, and change this to `self.workspace_metadata.update(workspace_metadata)`, then you can specify metadata when you write the step class (which is useful for things like the artifact type, which will always be the same), and to the step constructor/configuration file, which is useful for things like machine limits (because bigger models need more resources than smaller ones).

dirkgr (changes requested):
It doesn't just have to be for the workspace, right? It could just be "metadata".

tango/step.py
```diff
@@ -242,6 +251,9 @@ def __init__(
         ] = None  # This is set only while the run() method runs.
         self._config = step_config
         self.step_resources = step_resources
+        self.metadata = deepcopy(self.METADATA)
+        if step_metadata:
+            self.metadata.update()
```
dirkgr, line 256:
```suggestion
            self.metadata.update(step_metadata)
```

dirkgr (changes requested):

dirkgr (approved):

PR #364, opened 2022-08-20
Fix Trainer crash

dirkgr:


tango/integrations/torch/train.py
```diff
@@ -698,17 +698,21 @@ def save_state(step: int):
...
-                elif config.minimize_val_metric and val_metric <= best_val_metric:
-                    best_val_metric = val_metric
-                elif not config.minimize_val_metric and val_metric >= best_val_metric:
+                if (
+                    (best_val_metric is None)
+                    or (config.minimize_val_metric and val_metric <= best_val_metric)
+                    or (not config.minimize_val_metric and val_metric >= best_val_metric)
+                ):
                     best_val_metric = val_metric
+                    save_state(step)
```
epwalsh, line 707: I think what I said before on Slack was wrong: "We should definitely save state when there's a new best model"

Actually the intention of having separate `validate_every` and `checkpoint_every` arguments was to avoid the coupling of validating and checkpointing. It can be useful, for example, to checkpoint less often than you validate when validation is relatively cheap compared to checkpointing (often the case when you have a small validation set and a really big model). I had this exact situation when working on the staged training project.

That aside, another issue here is that we also call `save_state()` below on line 727, possibly duplicating the checkpointing work.

Here's an alternative suggestion, let me know if this would work:

When we catch a `StepIteration` error below we could check if there is a checkpoint, and if not, save one then. Assuming my understanding of the issue is correct...

epwalsh (changes requested):

dirkgr:
Closed in favor of #366.

PR #366, opened 2022-08-23
Fix undefined behavior with `TorchTrainStep`

epwalsh:
This would replace #364, though I think we should port the tests
over from that.

@dirkgr I think this addresses all of the edge cases we talked about.

### DIFF
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -485,10 +485,12 @@ def _train(
         the best checkpoint path to the current checkpoint file.
         """
         if val_metric is not None:
-            assert best_val_metric_checkpointed is not None
-            return (config.minimize_val_metric and val_metric <= best_val_metric_checkpointed) or (
-                not config.minimize_val_metric and val_metric >= best_val_metric_checkpointed
-            )
+            if best_val_metric_checkpointed is not None:
+                return (
+                    config.minimize_val_metric and val_metric <= best_val_metric_checkpointed
+                ) or (not config.minimize_val_metric and val_metric >= best_val_metric_checkpointed)
+            else:
+                return False
         else:
             # Without a validation loop we always treat the most recent checkpoint as the best.
             return True
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -268,6 +268,12 @@ class TorchTrainStep(Step):
                 "One of 'train_steps' or 'train_epochs' needs to be specified, but not both."
             )
 
+        if validate_every is not None and checkpoint_every is not None:
+            if checkpoint_every % validate_every != 0 and validate_every % checkpoint_every != 0:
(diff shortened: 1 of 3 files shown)


tango/integrations/torch/train.py
```diff
@@ -466,12 +487,12 @@ def is_best_checkpoint() -> bool:
...
-        if val_metric is not None and best_val_metric is not None:
-            return (config.minimize_val_metric and val_metric <= best_val_metric) or (
-                not config.minimize_val_metric and val_metric >= best_val_metric
+        if val_metric is not None and best_val_metric_checkpointed is not None:
+            return (config.minimize_val_metric and val_metric <= best_val_metric_checkpointed) or (
+                not config.minimize_val_metric and val_metric >= best_val_metric_checkpointed
             )
         elif best_batch_loss is not None:
-            return best_batch_loss <= batch_loss
+            return batch_loss <= best_batch_loss
```
dirkgr, line 495: I'm not sure we should keep the logic that keeps the best checkpoint based on batch loss. Batch loss is so noisy this is essentially random.

dirkgr: Maybe "total loss since the last time we checkpointed". That would be less random. But that feels like a whole new feature.

dirkgr: If we need some always-available fallback metric for model quality, I think recency will be better than "loss on the last batch before we checkpointed".

dirkgr (changes requested):
FWIW, I have hit this case now in a real scenario as well. I didn't debug deeply, but you can get there even without early stopping.

dirkgr:
I ported the test over.

tango/integrations/torch/train.py
epwalsh: Hmm yea using recency is probably okay. bfd336c

dirkgr: I don't think you pushed that.

epwalsh: Oops, yea push failed. Just pushed now

tango/integrations/torch/train.py
```diff
@@ -484,38 +481,35 @@ def _train(
 
     def is_best_checkpoint() -> bool:
         """
-        A closure that we'll call when saving checkpoints to check if we should hardlink
+        A closure that we'll call when saving checkpoints to check if we should link
```
dirkgr, line 484: To be sure, this has to be a hardlink, because the checkpoint that's saved under a name that derives from the step count might go away.

I guess we could do it the other way, where the current best checkpoint becomes a real file, with a symlink pointing to it. Then it doesn't have to be a hardlink anymore. But either way it doesn't work on Windows, so why bother?

dirkgr (approved):

dirkgr:
Argh, this is a real test failure now.

PR #367, opened 2022-08-24
Auto import local Python modules when searching for registered classes

epwalsh:

- When Tango is looking for a registered class (like a `Step`), it will now automatically search Python modules in the current working directory after unsuccessfully looking in all of the other usual spots. As a result, users don't necessarily need to specify the `--include-package` option.

I realized the need for this when giving the Tango demo last week. All of the sudden it seemed silly that I would need to always specify `--include-package steps.py` when Tango could just assume that I would want to include the Python files in my repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Tango will now automatically search Python modules in the current working directory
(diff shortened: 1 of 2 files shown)


dirkgr (approved):
This is cool, though doesn't Python natively prefer local files to everything else? So maybe resolving to local files should come first, to match what happens when you `import`?

epwalsh:
> maybe resolving to local files should come first, to match what happens when you import?

Yea that's a good point

PR #369, opened 2022-08-25
Detailed train outputs

lgatys:

- The `TorchTrainStep` now enables monitoring arbitrary model outputs during training.
- `TorchTrainEngine.forward_train` now returns a tuple `loss, model_outputs` for each micro batch.
- The list of model outputs for all micro batches in a batch is passed to the `TrainCallback.log_batch` and `TrainCallback.post_batch`.
- The documentation in the `log_batch` and `post_batch` functions  of the `TrainCallback` for why `batch_outputs` is `List[Dict[str, Any]]` is copied from the `pre_batch` function which has the same format for the `batch` input.

## Before submitting

## After submitting

### DIFF
--- a/tango/integrations/torch/train.py
+++ b/tango/integrations/torch/train.py
@@ -617,7 +617,10 @@ def _train(
                     raise ValueError("nan loss encountered")
                 batch_loss += micro_batch_loss.detach().item()
                 batch_outputs.append(
-                    {key: output.detach() if isinstance(output, torch.Tensor) else output for key, output in micro_batch_outputs.items()}
+                    {
+                        key: output.detach() if isinstance(output, torch.Tensor) else output
+                        for key, output in micro_batch_outputs.items()
+                    }
                 )
 
                 # Calculate gradients.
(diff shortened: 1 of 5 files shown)


tango/integrations/torch/train.py
```diff
@@ -607,25 +607,30 @@ def save_state(step: int):
...
                 # Get loss.
-                micro_batch_loss = training_engine.forward_train(
+                micro_batch_loss, micro_batch_outputs = training_engine.forward_train(
                     micro_batch, micro_batch_idx, len(batch)
                 )
                 if torch.isnan(micro_batch_loss):
                     raise ValueError("nan loss encountered")
                 batch_loss += micro_batch_loss.detach().item()
+                batch_outputs.append(
+                    {key: output.detach() for key, output in micro_batch_outputs.items()}
```
epwalsh, line 620: Good idea with detaching ✅

We should handle non-tensor outputs as well:

```suggestion
                    {key: output.detach() if isinstance(output, torch.Tensor) else output for key, output in micro_batch_outputs.items()}
```

epwalsh (changes requested):
Looks great! Just one minor change requested

lgatys:
> Looks great! Just one minor change requested

Thanks! I added your suggestion and fixed the tests. I think the remaining ones that are failing have to be re-run from your end?

epwalsh (approved):
LGTM!

PR #371, opened 2022-08-25
Allowing indexing into step results in config

### LINKED ISSUE #334: Support Indexing for multiple return values from a Tango Step

If a Tango step returns multiple values, we currently have to pass all the returned values to another Step. The goal is to support indexing so that the user can select which returned values should be passed as an input to another Step. 

### Alternatives

_No response_

### Additional context

_No response_

epwalsh:
For example, ref steps can now look like
`{type: "ref", ref: some_step, key: 0}`

Closes #334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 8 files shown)


dirkgr (changes requested):
I'm pretty sure that case needs to be handled in `Step._ordered_dependencies()`. The fact that this didn't become apparent in a test means we need a test for that.

Are there other places where we have a big `if ... elif` ladder involving `Step`?

epwalsh:
@dirkgr yeup, good catch. Pretty sure I got them all now: b331b26

dirkgr (approved):

PR #373, opened 2022-08-30
Only the Beaker executor needs clean working directories

dirkgr:


tango/integrations/beaker/executor.py
```diff
@@ -290,6 +291,14 @@ def execute_step_graph(
     ) -> ExecutorOutput:
         import concurrent.futures
 
+        # Make sure repository is clean, if we're in one.
+        try:
+            repo = Repo(".")
+            if repo.is_dirty():
+                raise ValueError("You have uncommitted changes! Use --allow-dirty to force.")
```
epwalsh, line 298: There should be a way to circumvent this, but you've taken away the `--allow-dirty` flag

epwalsh (changes requested):

tango/integrations/beaker/executor.py
dirkgr: I thought the Beaker Executor will never do the right thing if your working directory is dirty, so there is no point in overriding the setting.

epwalsh: If you've changed a file that doesn't need to be committed for the Beaker executor to work, it will work fine.

In general though, having a uncommitted changes means the commit hash tied to `StepInfo` will be wrong. That's not just an issue with the Beaker executor. That's why I added that behavior in the first place.

dirkgr: This is why I think `StepInfo` is of limited utility. When I'm developing a step from scratch, I know I can just hit "run" repeatedly, as long as the step never succeeds.

epwalsh: Not sure what you're saying, but what I'm suggesting is:
1. Add an option to the `BeakerExecution` to allow uncommitted changes, and
2. Update this error message:
```suggestion
                raise ValueError("You have uncommitted changes! Use the 'allow_dirty' option to force.")
```

dirkgr: I'm working on this.

What happens when I have committed changes that I have not pushed?

epwalsh: We should definitely check for that as well if possible. I couldn't figure out a good way to do that before, but I'll take a 2nd look

epwalsh: After reading through the answers to https://stackoverflow.com/questions/2016901/viewing-unpushed-git-commits, I think this would work:

```
git log origin..HEAD --not --remotes --oneline
```

(assuming "origin" is the name of the remote we're using)

Not sure yet if we can do this through the git Python client.

epwalsh: Ah ha!

```python
from git import Git
git = Git(".")
print(git.log(["origin..HEAD", "--not", "--remotes", "--oneline"]))
```

epwalsh (approved):
LGTM

PR #374, opened 2022-08-30
Add `Workspace.step_result()` method

epwalsh:

- Adds a method for getting the latest result for a given step. This is usually more convenient than `Workspace.step_result_for_run()`, which requires knowing the run name.


dirkgr (approved):
This encourages a messy and confusing way of working with Tango. But I'm trying not to be dogmatic about it.

PR #379, opened 2022-09-01
Catch all errors when attempting log record decode

epwalsh:
Or, as @dirkgr suggested, we could just not stream the logs. But I think we should merge this either way so that we have this fix *for the record*.

### DIFF
--- a/tango/integrations/beaker/executor.py
+++ b/tango/integrations/beaker/executor.py
@@ -460,7 +460,8 @@ class BeakerExecutor(Executor):
                         # We don't need to handle logs from Tqdm since that would result in
                         # duplicate messages (those Tqdm lines get printed directly to the output as well).
                         logging.getLogger(log_record.name).handle(log_record)
-                except JSONDecodeError:
+                except:  # noqa: E722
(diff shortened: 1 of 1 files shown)


tango/integrations/beaker/executor.py
```diff
@@ -460,7 +460,8 @@ def _execute_sub_graph_for_step(
                         # We don't need to handle logs from Tqdm since that would result in
                         # duplicate messages (those Tqdm lines get printed directly to the output as well).
                         logging.getLogger(log_record.name).handle(log_record)
-                except JSONDecodeError:
+                except:  # noqa: E722
```
dirkgr, line 463: I think at the very least we should catch `Exception`, so that `KeyboardInterrupt`, `OutOfMemory`, and stuff like that doesn't get swallowed up.

dirkgr (changes requested):

tango/integrations/beaker/executor.py
epwalsh: Good pointd: 4089ac

PR #392, opened 2022-09-07
Improve Beaker cluster choice logic

epwalsh:


tango/integrations/beaker/executor.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import os
+import random
```
dirkgr, line 3: This is unused?

tango/integrations/beaker/executor.py
```diff
@@ -590,18 +592,32 @@ def _ensure_step_graph_dataset(self, step_graph: StepGraph) -> Dataset:
...
+            def recency_ranking(cluster_name: str):
+                try:
+                    return self._latest_clusters_used.index(cluster_name)
+                except ValueError:
+                    return -1
+
             available_clusters = sorted(
                 self.beaker.cluster.filter_available(task_resources, *self.clusters),
-                key=lambda x: x.queued_jobs,
+                key=lambda x: (x.queued_jobs, recency_ranking(x.cluster.full_name)),
```
dirkgr, line 610: `queued_jobs` is the number of jobs queued but not running? Does it update every time you call it?

epwalsh: Oops. Yes that does update every time

dirkgr: It seems to me that this should sort by number of free slots first ... but I guess it's hard to find out what a "free slot" is given resource requirements.

As a stop-gap until Beaker has multi-cluster support, this is great.

dirkgr (approved):

PR #397, opened 2022-09-09
Make `BeakerExecutor` more robust to all recoverable errors types (connection, HTTP, SSL, timeout, etc)

epwalsh:
Followup to https://github.com/allenai/beaker-py/pull/152

### DIFF
--- a/tango/integrations/beaker/workspace.py
+++ b/tango/integrations/beaker/workspace.py
@@ -94,9 +94,7 @@ class BeakerWorkspace(Workspace):
             if file_info.digest in self._step_info_cache:
                 step_info = self._step_info_cache.pop(file_info.digest)
             else:
-                step_info_bytes = b"".join(
-                    self.beaker.dataset.stream_file(dataset, file_info, quiet=True)
-                )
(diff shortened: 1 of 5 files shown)


dirkgr (approved):

tango/integrations/beaker/workspace.py
```diff
@@ -291,8 +289,8 @@ def _get_run_from_dataset(self, dataset: Dataset) -> Optional[Run]:
         try:
             run_name = dataset.name[len(Constants.RUN_DATASET_PREFIX) :]
             steps: Dict[str, StepInfo] = {}
-            steps_info_bytes = b"".join(
-                list(self.beaker.dataset.stream_file(dataset, Constants.RUN_DATA_FNAME, quiet=True))
+            steps_info_bytes = self.beaker.dataset.get_file(
```
epwalsh, line 292: `get_file()` is a little more robust than `stream_file()` because it will treat `ChecksumFailed` as a recoverable error and retry the whole download. That's not possible with `stream_file()`.

PR #398, opened 2022-09-09
Deterministic hashing for tensors

dirkgr:


tango/common/det_hash.py
```diff
@@ -82,9 +94,12 @@ def det_hash_object(self) -> Any:
             return None  # When you return `None` from here, it falls back to just hashing the object itself.
 
 
+_PICKLE_PROTOCOL = 4
```
AkshitaB, line 97: Can probably select protocol 5, since we require the minimum Python version to be 3.8.

AkshitaB (approved):
LGTM. Minor comment about the pickle protocol.

dirkgr:
Hmm, that would break all backwards compatibility though. All hashes of
existing projects would change.

PR #400, opened 2022-09-13
Fix bug with `StepInfo.from_params()`, canceling `BeakerExecutor`, reserve "ref" name

epwalsh:
- Fixes another bug with `StepInfo.from_params()` which would result in `from_params` / `Registrable` trying to instantiate things they shouldn't be. Normally this isn't a big deal because the in the end everything gets created how it should, but it slows things down substantially when you have a large step graph. This is related to #344.
- Fixes a bug with the `BeakerExecutor` where hitting CTRL-C doesn't work as expected. It just hangs instead of stopping the experiments that were already submitted. For some reason I only ran into this bug when running a large step graph.
- Makes the `BeakerWorkspace` more robust to finding step info data in an invalid state, which can happen quite often when Beaker jobs are stopped prematurely.
- 🚧  Safety improvement that I thought of when I was fixing the first bug: We should make sure "ref" is a reserved name that can't be used to register a step.

### DIFF
--- a/tango/common/registrable.py
+++ b/tango/common/registrable.py
@@ -123,6 +123,13 @@ class Registrable(FromParams):
                     ...  # construct some_params from files
                     return cls(some_params)
         """
+        from tango.step import Step
+
+        if cls == Step and name == "ref":
+            raise ConfigurationError(
+                "You cannot use the name 'ref' to register a step. This name is reserved."
+            )
+
         registry = Registrable._registry[cls]
 
         def add_subclass_to_registry(subclass: Type[_T]) -> Type[_T]:
@@ -178,7 +185,12 @@ class Registrable(FromParams):
         """
         Search for and import modules where ``name`` might be registered.
         """
+        from tango.step import Step
+
         if could_be_class_name(name) or name in Registrable._registry[cls]:
+            return None
+
+        if cls == Step and name == "ref":
             return None
 
         def try_import(module):
--- a/tango/common/registrable.py
+++ b/tango/common/registrable.py
@@ -123,9 +123,8 @@ class Registrable(FromParams):
                     ...  # construct some_params from files
                     return cls(some_params)
         """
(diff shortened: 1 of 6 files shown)


tango/common/registrable.py
```diff
@@ -340,3 +347,14 @@ def _get_suggestion(name: str, available: List[str]) -> Optional[str]:
         if suggestion in available:
             return suggestion
     return None
+
+
+def _fullname(c: type) -> str:
+    return f"{c.__module__}.{c.__qualname__}"
```
dirkgr, line 353: Is this about making pickle work?

tango/integrations/beaker/workspace.py
```diff
@@ -118,7 +119,16 @@ def step_starting(self, step: Step) -> None:
         self.locks[step] = lock
 
         step_info = self.step_info(step)
-        if step_info.state not in {StepState.INCOMPLETE, StepState.FAILED, StepState.UNCACHEABLE}:
+        if step_info.state == StepState.RUNNING:
+            # Since we've acquired the step lock we know this step can't be running
+            # elsewhere. But the step state can still say its running if the last
+            # run exited before this workspace had a chance to update the step info.
```
dirkgr, line 125: The `LocalWorkspace` has a function that guesses the step state given what we know about locks, existing directories, etc. Maybe it would be helpful to steal from there.

dirkgr (approved):

tango/common/registrable.py
epwalsh: No, I just pulled this function out to the module level (previously it was just a locally defined function in `.register()`.

tango/integrations/beaker/workspace.py
epwalsh: I forgot about that. But I guess this is a simplified version because I think there's less edge cases to check here.

PR #401, opened 2022-09-13
Make `BeakerStepLock` more robust

epwalsh:
Makes the `BeakerStepLock` more robust to being left in a bad state, and as a result the `BeakerWorkspace` is more robust.

In particular, there are two changes here that make this work:
1. The `BeakerStepLock` is made aware of the Beaker experiment that last acquired the lock (if there was one). So if the lock is left in a bad state by that Beaker experiment, another process trying to acquire the lock can just check the state of the experiment.
2. The `BeakerStepLock` registers an `atexit` hook to call `release()`. This makes it way less likely that the lock will be left in a bad state.

### DIFF
--- a/tango/integrations/beaker/common.py
+++ b/tango/integrations/beaker/common.py
@@ -1,9 +1,20 @@
+import atexit
+import json
 import logging
+import tempfile
 import time
 import urllib.parse
-from typing import Optional, Union
+from pathlib import Path
+from typing import Any, Dict, Optional, Union
 
-from beaker import Beaker, Dataset, DatasetConflict, DatasetNotFound
+from beaker import (
+    Beaker,
+    Dataset,
+    DatasetConflict,
+    DatasetNotFound,
+    Experiment,
+    ExperimentNotFound,
+)
 
 from tango.step import Step
 from tango.step_info import StepInfo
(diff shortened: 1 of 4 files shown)


tango/integrations/beaker/common.py
```diff
@@ -95,6 +172,7 @@ def release(self):
                 # Dataset must have been manually deleted.
                 pass
             self._lock_dataset = None
+            atexit.unregister(self.release)
```
dirkgr, line 175: `atexit` works properly with methods?

dirkgr (approved):
This is our first "better together" feature 😶 .

tango/integrations/beaker/common.py
epwalsh: Apparently! Although one side-effect is that if you have an acquired lock and call `del lock`, the `__del__` method won't actually be called immediately since a reference to the lock will still exist in the `atexit` hook. That's not ideal, and wouldn't be okay IMO if `BeakerStepLock` was meant to be used outside of Tango. But since it's only used internally I think that's okay. Just means we have to explicitly call `lock.release()` instead of relying on `lock.__del__()`.

PR #405, opened 2022-09-14
Make GPU CI tests preemptible on Beaker

epwalsh:
As of https://github.com/allenai/beaker-run-action/pull/37 this should work seamlessly, and will make us better Beaker citizens.


dirkgr (approved):
I hope you're right. First time I used preemptible Beaker jobs it didn't work right.

dirkgr:
I don't understand how this can happen, but it seems this makes GPU tests fail?

epwalsh:
Ugh, this is a bug. I'm going to close until I fix this on beaker-run-action.

PR #406, opened 2022-09-14
Ensure base path for included modules kept in `sys.path`

### LINKED ISSUE #404: ModuleNotFoundError when using multiprocessing in a Step
### 🐛 Describe the bug

I encountered **ModuleNotFoundError** when I try to use multiprocessing in a Step. However, when I run the same Step without using tango, everything is ok.

**To replicate**:

directory structure:
```
- test_tango
    - __init__.py
    - my_step.py
```

my_step.py:
```python
from joblib import delayed, Parallel
from tango import Step

def add_one(a):
    return a+1

@Step.register("test1")
class Test1(Step):
    def run(self):
        tasks = [delayed(add_one)(i) for i in range(10)]
        results = Parallel(n_jobs=-1, prefer="processes")(tasks)
        print(results)
```

test_tango.jsonnet:
```jsonnet
{
    "steps": {
        "step1": {
             "type": "test1"
        }
    }
}
```

**Run with tango**:
```bash
tango run test_tango.jsonnet -i test_tango -w workspace
```

got:
```
ERROR    [step step1] Uncaught exception                                                                                                 logging.py:373
                           joblib.externals.loky.process_executor._RemoteTraceback:
                           """
                             File "$PATH/lib/python3.9/site-packages/joblib/externals/loky/process_executor.py", line
                           407, in _process_worker
                               call_item = call_queue.get(block=True, timeout=timeout)
                             File "$PATH/lib/python3.9/multiprocessing/queues.py", line 122, in get
                               return _ForkingPickler.loads(res)
                           ModuleNotFoundError: No module named 'test_tango'
```

**Directly use the Step**:
```python
>>> from test_tango.my_step import Test1
>>> runner = Test1()
>>> runner.run()
[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

Everything is ok.

### Versions

Python 3.9.12
ai2-tango==0.11.0

epwalsh:
Fixes #404. I'm not sure why we didn't keep the path in `sys.path` before.. probably just to avoid potential side-effects?

### DIFF
--- a/tango/common/util.py
+++ b/tango/common/util.py
@@ -5,7 +5,6 @@ import string
 import sys
 import traceback
 from collections import OrderedDict
-from contextlib import contextmanager
 from dataclasses import asdict, is_dataclass
 from datetime import datetime, tzinfo
 from pathlib import Path
@@ -13,7 +12,6 @@ from typing import Any, Iterable, Optional, Set, Tuple, Union
 
 import pytz
 
-from .aliases import PathOrStr
 from .exceptions import SigTermReceived
 
 
@@ -33,25 +31,6 @@ def _handle_sigterm(sig, frame):
 
 def install_sigterm_handler():
     signal.signal(signal.SIGTERM, _handle_sigterm)
-
-
-@contextmanager
-def push_python_path(path: PathOrStr):
-    """
-    Prepends the given path to `sys.path`.
-
-    This method is intended to use with `with`, so after its usage, its value willbe removed from
-    `sys.path`.
-    """
-    # In some environments, such as TC, it fails when sys.path contains a relative path, such as ".".
-    path = Path(path).resolve()
-    path = str(path)
-    sys.path.insert(0, path)
-    try:
-        yield
-    finally:
-        # Better to remove by value, in case `sys.path` was manipulated in between.
-        sys.path.remove(path)
 
 
 _extra_imported_modules: Set[str] = set()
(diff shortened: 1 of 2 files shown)


epwalsh:
>  I'm not sure why we didn't keep the path in sys.path before.. probably just to avoid potential side-effects?

Looking back, seems like it was initially implemented in AllenNLP with the idea that people might want to "unload" plugins:
https://github.com/allenai/allennlp/pull/3671#issuecomment-578291461

tango/common/util.py
```diff
@@ -108,29 +108,32 @@ def import_module_and_submodules(package_name: str, exclude: Optional[Set[str]]
...
 
     if exclude and package_name in exclude:
         return
 
     importlib.invalidate_caches()
 
-    # For some reason, python doesn't always add this by default to your path, but you pretty much
-    # always want it when using `--include-package`.  And if it's already there, adding it again at
-    # the end won't hurt anything.
-    with push_python_path(base_path):
```
AkshitaB, line 120: @epwalsh I thought `push_python_path` is basically doing this with the context manager. But for some reason that doesn't work with multiprocessing?

AkshitaB: We can remove `push_python_path` from the code then, since it's not being used elsewhere.

AkshitaB (changes requested):
If

tango/common/util.py
epwalsh: `push_python_path` temporarily adds the right path to `sys.path`. The problem is in this case is that the multiprocessing library @Tbabm is using needs to re-import some modules to deserialize objects, and at that point the needed path is no longer on `sys.path`.

epwalsh: Removed: b16315c

AkshitaB: Got it.

AkshitaB (approved):

PR #407, opened 2022-09-14
Added `BeakerScheduler` class for handling resource assignment

epwalsh:
This allows users to implement their own subclasses to customize how the `BeakerExecutor` allocates resources to run each step. Here's an example of a custom implementation that makes jobs preemptible if they're not run on AllenNLP clusters:

https://github.com/allenai/tango-beaker-template/blob/0346e8719388cf8f4cc0c80ac713ae14f570f7e0/scheduler.py

### DIFF
--- a/tango/integrations/beaker/executor.py
+++ b/tango/integrations/beaker/executor.py
@@ -140,13 +140,15 @@ class SimpleBeakerScheduler(BeakerScheduler):
         )
         cluster_to_use = self._ensure_cluster(task_resources)
         if cluster_to_use is None:
-            cluster_to_use = self.clusters[0]
+            raise ResourceAssignmentError()
+
         # Move cluster to the end of `self._latest_clusters_used`
         try:
             self._latest_clusters_used.remove(cluster_to_use)  # type: ignore
         except ValueError:
             pass
         self._latest_clusters_used.append(cluster_to_use)  # type: ignore
+
         return ResourceAssignment(
             cluster=cluster_to_use, resources=task_resources, priority=self.priority
         )
--- a/tango/integrations/beaker/executor.py
+++ b/tango/integrations/beaker/executor.py
@@ -53,6 +53,12 @@ class StepFailedError(ExecutorError):
         self.experiment_url = experiment_url
 
 
+class ResourceAssignmentError(ExecutorError):
+    """
+    Raised when a scheduler can't find enough free resources at the moment to run a step.
+    """
+
+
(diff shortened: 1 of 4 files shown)


tango/integrations/beaker/executor.py
```diff
@@ -49,6 +53,120 @@ def __init__(self, msg: str, experiment_url: str):
...
+            cluster_to_use = self.clusters[0]
+        # Move cluster to the end of `self._latest_clusters_used`
+        try:
+            self._latest_clusters_used.remove(cluster_to_use)  # type: ignore
+        except ValueError:
+            pass
+        self._latest_clusters_used.append(cluster_to_use)  # type: ignore
+        return ResourceAssignment(
+            cluster=cluster_to_use, resources=task_resources, priority=self.priority
+        )
```
dirkgr, line 143: Is this method allowed to return a value that says "We can't schedule this one right now."? I'm thinking of a scheduler that doesn't queue more than a few jobs ahead of time, and then waits to see which cluster frees up first. Though once again that's a problem that goes away when Beaker fixes https://github.com/allenai/beaker/issues/2544.

dirkgr (approved):

tango/integrations/beaker/executor.py
epwalsh: At the moment, no. But that would be nice. It would take some refactoring though so should probably be a separate PR.

epwalsh: Oh! Just kidding, that was easy: 76e5e4d

epwalsh: And cb438bd

tango/integrations/beaker/executor.py
```diff
@@ -533,16 +542,23 @@ def future_done_callback(future: concurrent.futures.Future):
                             result_location=self.workspace.step_info(step).result_location,
                             logs_location=future.result(),
                         )
+                        steps_left_to_run.discard(step)
+                    elif isinstance(exc, ResourceAssignmentError):
+                        submitted_steps.discard(step_name)
```
dirkgr, line 549: This doesn't mean they are discarded forever, does it?

dirkgr (approved):

tango/integrations/beaker/executor.py
epwalsh: Nope, they get put back into the queue, or I guess _kept_ in the queue. The important queue here is `steps_left_to_run`.

PR #414, opened 2022-09-20
`BeakerExecutor` won't run duplicate cacheable steps

### LINKED ISSUE #412: Beaker Executor should not run steps it's already running

@dirkgr: When it detects that a step it needs is already running (but it’s not itself running it), it should not schedule that step, but still wait for its completion. Maybe print a link every once in a while so the user can check and make sure it’s really running.

### Alternatives

Just let the workspace handle that. Workspaces do already have a mechanism for not running multiple of the same cacheable step at once. But the Beaker executor will still happily submit experiments for the same cacheable step, which can waste a lot of time.

### Additional context

We can implement this in the Beaker executor by searching for experiments in the workspace that match the name of the step. Then just check the status of the most recent matching experiment.

epwalsh:
Fixes #412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where canceling the Beaker executor might not work properly.
 - Fixed a bug where the trainer trains too much when `train_epochs` is set and you're using gradient accumulation.
 - Fixed how the results of uncacheable steps are displayed by `tango run`.
+- Beaker executor won't run duplicate cacheable steps at the same time.
 
 ## [v0.13.0](https://github.com/allenai/tango/releases/tag/v0.13.0) - 2022-09-07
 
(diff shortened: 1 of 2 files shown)


dirkgr (approved):
This is fantastic. It means I can slowly build up my large experiment without having to think about whether some new run overlaps with a run already in progress.

Let's say I have two runs going that include the same step. Run 1 decides to kick off the step. What happens when
1. The step completes, and later run 2 also wants to run the step?
2. The step hasn't completed yet, and run 2 wants to run the step?

epwalsh:
> What happens when
>
> 1. The step completes, and later run 2 also wants to run the step?
> 2. The step hasn't completed yet, and run 2 wants to run the step?

In (1), run 2 will find the result of the step in the cache when it's trying to figure out if it should should run or not, and so won't run it.

In (2), run 2 will find the existing experiment for the step, and see that it's still running, so will just follow that experiment instead of submitting a new one.

PR #417, opened 2022-09-22
GCSWorkspace

AkshitaB:

* Adds the following abstractions: `RemoteClient`, `RemoteStepCache`, `RemoteWorkspace`. The `RemoteClient` models the beaker client (in fact, for beaker, it's mostly a thin wrapper right now).
* GCS and Beaker step caches and workspaces inherit from above.
* WandBWorkspace is too different to easily fit into the RemoteWorkspace abstractions, leaving that refactoring for a later date.

Cleanup:
- [x] Address TODOs
- [x] Test with larger config.
- [x] Test BeakerExecutor with GCSWorkspace.
- [x] Add docstrings

**Testing with BeakerExecutor:**

* Created from the tango-beaker-template: https://github.com/allenai/tango-beaker-gcs-testing (Currently using this branch).
* Just change the workspace details to use GS:
```
workspace:
  type: "gs"
  workspace: "allennlp-tango-bucket"
```


tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,159 @@
...
+
+# This class inherits from `LocalStepCache` to benefit from its in-memory "weak cache" and "strong cache",
+# but it handles saving artifacts to disk a little differently.
+class RemoteStepCache(LocalStepCache):
+    """
+    This is a :class:`~tango.step_cache.StepCache` that's used by :class:`RemoteWorkspace`.
+    It stores the results of steps on some RemoteWorkspace.
+
+    It also keeps a limited in-memory cache as well as a local backup on disk, so fetching a
+    step's resulting subsequent times should be fast.
```
epwalsh, line 35:
```suggestion
    step's result subsequent times should be fast.
```

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,159 @@
...
+            return False
+
+    def _step_results_dir(self) -> str:
+        return RemoteConstants.STEP_RESULT_DIR
+
+    def _step_result_remote(self, step: Union[Step, StepInfo]):
+        raise NotImplementedError()
+
+    def _fetch_step_remote(self, step_result, target_dir: PathOrStr):
+        raise NotImplementedError()
```
epwalsh, line 82: I think it would be good to annotate the return type of these methods, and also decorate them with `@abstractmethod` (from the `abc` module)

tango/workspaces/remote_workspace.py
```diff
@@ -0,0 +1,344 @@
...
+                )
+            except RemoteDatasetConflict:
+                raise ValueError(f"Run name '{name}' is already in use")
+
+        steps: Dict[str, StepInfo] = {}
+        run_data: Dict[str, str] = {}
+
+        # Collect step info.
+        with concurrent.futures.ThreadPoolExecutor(
+            thread_name_prefix="GCSWorkspace.register_run()-"
```
epwalsh, line 230: Maybe something more generic?
```suggestion
            thread_name_prefix="RemoteWorkspace.register_run()-"
```

epwalsh (commented):
I like the new abstractions. My main question is about remote locking. It seems like GCSFs does have the notion of "committing" a file, similar to Beaker (https://gcsfs.readthedocs.io/en/latest/api.html#gcsfs.core.GCSFile.commit), but it doesn't look like we use that feature? From what I could tell, the "locking" mechanism in `GCSClient` works by checking for the existence of some placeholder file? Does [`gcs_fs.touch(placeholder_file)`](https://github.com/allenai/tango/pull/417/files#diff-9ecdbe82049ab6590ac044a287632dc6bd1625c56059083c74afc51d73f0d894R113) fail if another process beats it by touching that file first?

General comments:
- I think `tango/integrations/(beaker|gs)/common.py` and `...util.py` can be merged into the same file.

AkshitaB:
> My main question is about remote locking. It seems like GCSFs does have the notion of "committing" a file, similar to Beaker (https://gcsfs.readthedocs.io/en/latest/api.html#gcsfs.core.GCSFile.commit), but it doesn't look like we use that feature? 
That commit() works for files rather than folders. But perhaps I can utilize it recursively (Looking into it). From what I understand, beaker also performs commit at the folder level, right?

> From what I could tell, the "locking" mechanism in `GCSClient` works by checking for the existence of some placeholder file? 
The placeholder file is for creation of any folder, since you can't create an empty folder in gcs.

> * I think `tango/integrations/(beaker|gs)/common.py` and `...util.py` can be merged into the same file.

Agreed.

dirkgr:
Oh wow! https://cloud.google.com/storage/docs/consistency
They do promise consistency for file listings. So lock files should work. That's way better than S3!

epwalsh:
> From what I understand, beaker also performs commit at the folder level, right?

At the "dataset" level, which is similar I guess

dirkgr:
This is ready for more reviewing?

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,232 @@
...
+            path = os.path.join(self.bucket_name, dataset)
+        else:
+            # We have a dataset, and we recreate it with refreshed info.
+            path = dataset.dataset_path
+        try:
+            return self._convert_ls_info_to_dataset(self.gcs_fs.ls(path=path, detail=True))
+        except FileNotFoundError:
+            raise RemoteDatasetNotFound()
+
+    def create(self, dataset: str, commit: bool = False):
```
epwalsh, line 114: It doesn't look like the `commit` parameter is used here. Should it be?

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,232 @@
...
+    def fetch(self, dataset: GCSDataset, target_dir: PathOrStr):
+        try:
+            self.gcs_fs.get(dataset.dataset_path, target_dir, recursive=True)
+        except FileNotFoundError:
+            raise RemoteDatasetNotFound()
+
+    def datasets(
+        self, match: str, uncommitted: bool = True, results: bool = False
+    ) -> List[GCSDataset]:
+        # TODO: remove "results", it's meant to be beaker-specific?
```
epwalsh, line 202: Yea, that Beaker specific and probably doesn't need to be part of this API.

epwalsh (commented):
> WandBWorkspace is too different to easily fit into the RemoteWorkspace abstractions, leaving that refactoring for a later date.

Yea I don't think we should worry about this since we've been talking about removing that workspace from the core repo anyway.

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,199 @@
...
+            self.step_dir(step).with_suffix(".lock"), read_only_ok=read_only_ok
+        ).acquire_with_updates(desc=f"acquiring step cache lock for '{step.unique_id}'")
+
+    def __contains__(self, step: Any) -> bool:
+        if isinstance(step, (Step, StepInfo)):
+            cacheable = step.cache_results if isinstance(step, Step) else step.cacheable
+            if not cacheable:
+                return False
+
+            # TODO: old beaker step cache seems to not check locally. Should it?
```
dirkgr, line 115: Yes, but out of scope for this PR.

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def __len__(self):
+        # NOTE: lock datasets should not count here. They start with the same prefix,
+        # but they never get committed.
+        return sum(
+            1
+            for ds in self.client.datasets(
+                uncommitted=False, match=self.Constants.STEP_DATASET_PREFIX
+            )
+            if ds.name is not None and ds.name.startswith(self.Constants.STEP_DATASET_PREFIX)
```
dirkgr, line 95: Can `name` ever be `None`?

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+
+    def _step_results_dir(self) -> str:
+        return self.Constants.STEP_RESULT_DIR
+
+    def _raise_remote_not_found(self):
+        raise RemoteNotFoundError()
```
dirkgr, line 102: What are these for? The functions do almost nothing, and don't add any clarity to the code either.

tango/step_caches/remote_step_cache.py
dirkgr: To be clear, they all should check locally as an optimization, but adding that to Beaker is out of scope.

tango/workspaces/remote_workspace.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+
+    Constants = RemoteConstants
+    STEP_INFO_CACHE_SIZE = 512
+
+    def __init__(self):
+        super().__init__()
+
+        # TODO: remove after conferring.
+        # We make all this into properties because FromParams cannot handle the following construct.
```
dirkgr, line 51: What does `FromParams` say about the construct?

tango/workspaces/remote_workspace.py
```diff
@@ -0,0 +1,392 @@
...
+            steps: Dict[str, StepInfo] = {}
+            steps_info_bytes = self.client.get_file(dataset, self.Constants.RUN_DATA_FNAME)
+            steps_info = json.loads(steps_info_bytes)
+        except (RemoteDatasetNotFound, FileNotFoundError):
+            return None
+
+        import concurrent.futures
+
+        with concurrent.futures.ThreadPoolExecutor(
+            max_workers=9, thread_name_prefix="RemoteWorkspace._get_run_from_dataset()-"
```
dirkgr, line 366: This number 9 is carefully chosen to not overwhelm Beaker. It should not be hardcoded here. Google can likely handle more traffic.

tango/common/remote_utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+class RemoteDatasetWriteError(TangoError):
+    """
+    Error denoting that there was an issue writing the dataset to its remote location.
+    """
+
+    pass
+
+
+class RemoteClient(Registrable):
```
dirkgr, line 126: It seems to me like this is one level too many as far as abstractions go. There is a lot of commonality between `BeakerWorkspace` and `GCSWorkspace`, and you don't want to copy and paste that, so I see how you got here. Can you achieve the same thing by having some "protected" abstract class members in `RemoteWorkspace` that are overridden in `BeakerWorkspace` and `GCSWorkspace`, so we don't need to build up this parallel hierarchy of `*Client` classes?

tango/common/remote_utils.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        Returns the url of the dataset within the workspace.
+        """
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def full_name(self) -> str:
+        """
+        Returns the name of the workspace attached to the client.
```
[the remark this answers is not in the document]
dirkgr: Well, we already have a `RemoteWorkspace`, so probably not that name ...

With a name like "Client" I would expect that multiple workspaces can at least share one client. But that doesn't seem to be the case.

dirkgr (changes requested):

tango/integrations/beaker/executor.py
```diff
@@ -436,6 +440,17 @@ def __init__(
                 "or as the environment variable 'GITHUB_TOKEN'."
             )
 
+        self.google_token = google_token or os.environ.get("GOOGLE_TOKEN")
+        # If the user passes file path instead of actual credentials dict, load the credentials
+        # so that they can be used in beaker.
```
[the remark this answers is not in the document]
AkshitaB: This would be the scenario if someone used gcloud auth locally, and then runs stuff on beaker. We can insist that people manually copy the contents of the file into a string, and add that as an environment variable. Just thought that this would be convenient. Passing the filepath as token does work locally with gcsfs, so this extends that to `BeakerExecutor`.

tango/integrations/beaker/executor.py
```diff
@@ -219,6 +219,8 @@ class BeakerExecutor(Executor):
     :param beaker_workspace: The name or ID of the Beaker workspace to use.
     :param github_token: You can use this parameter to set a GitHub personal access token instead of using
         the ``GITHUB_TOKEN`` environment variable.
+    :param google_token: You can use this parameter to set a Google Cloud token instead of using
+        the ``GOOGLE_TOKEN`` environment variable.
```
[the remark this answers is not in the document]
AkshitaB: Yeah, it wasn't the most elegant solution, but since we expect to use gcsworkspace with beaker more, I'm ok with it.

tango/integrations/beaker/workspace.py
```diff
@@ -60,29 +24,50 @@
...
+    def locks(self):
+        return self._locks
+
+    @property
+    def steps_dir_name(self):
+        return "beaker_workspace"
+
+    @property
+    def beaker(self):
+        return self.client.beaker
```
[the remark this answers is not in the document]
AkshitaB: Adding them as property because of the same `FromParams` issue with arbitrary kwargs.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    A client for interacting with Google Cloud Storage.
+    """
+
+    placeholder_file = ".placeholder"
+    uncommitted_file = ".uncommitted"
+
+    def __init__(self, bucket_name: str, token: str = "google_default"):
+        # "Bucket names reside in a single namespace that is shared by all Cloud Storage users" from
+        # https://cloud.google.com/storage/docs/buckets. So, the project name does not matter.
```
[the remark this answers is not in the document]
AkshitaB: Google cloud project name. Typically, when you run gcloud auth, you also specify the project name -- this is relevant for deciding which project's VMs etc to use. But all buckets reside in the same place regardless of the project.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,230 @@
...
+
+class GCSClient(RemoteClient):
+    """
+    A client for interacting with Google Cloud Storage.
+    """
+
+    placeholder_file = ".placeholder"
+    uncommitted_file = ".uncommitted"
+
+    def __init__(self, bucket_name: str, token: str = "google_default"):
```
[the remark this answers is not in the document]
AkshitaB: Should be easy.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,230 @@
...
+        # TODO: potentially consider gsutil rsync with --delete --ignore-existing command
+        # Using gsutil programmatically:
+        # https://github.com/GoogleCloudPlatform/gsutil/blob/84aa9af730fe3fa1307acc1ab95aec684d127152/gslib/tests/test_rsync.py
+        try:
+            source = str(objects_dir)
+            if objects_dir.is_dir():
+                source += "/*"
+            for path in glob.glob(source):
+                basepath = os.path.basename(path)
+                self.gcs_fs.put(path, os.path.join(folder_path, basepath), recursive=True)
```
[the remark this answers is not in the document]
AkshitaB: The trouble is that that google buckets cannot handle the concept of an empty folder -- you can't create it programmatically. The file structure is flat. That's why I use a placeholder file.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,230 @@
...
+        else:
+            full_file_path = os.path.join(dataset.dataset_path, file_path)
+        with self.gcs_fs.open(full_file_path, "r") as file_ref:
+            file_contents = file_ref.read()
+        return file_contents
+
+    def file_info(self, dataset: GCSDataset, file_path: str) -> FileInfo:
+        full_file_path = os.path.join(dataset.dataset_path, file_path)
+        info = self.gcs_fs.ls(full_file_path, detail=True)[0]
+        # TODO: should digest be crc32c instead of md5Hash?
```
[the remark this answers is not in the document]
AkshitaB: Both are part of the info, so we aren't using more resources. I'll stick to md5.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,230 @@
...
+            dataset = self._convert_ls_info_to_dataset(info)
+            if not uncommitted:
+                if not dataset.committed:
+                    continue
+            list_of_datasets.append(dataset)
+        return list_of_datasets
+
+
+def _is_json_str(string: str) -> bool:
+    return "{" in string
```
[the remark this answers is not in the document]
AkshitaB: Yeah, not the best bit of code. I wanted something quick and dirty to test things. 🙈

tango/step_caches/remote_step_cache.py
AkshitaB: Well, actually, with these abstractions, beaker's step cache does check locally now.

tango/step_caches/remote_step_cache.py
AkshitaB: Yeah, leftover from when I was trying to build the abstractions.

tango/workspaces/remote_workspace.py
AkshitaB: `TypeError: from_params() got an unexpected keyword argument`. Because if it's a keyword argument, `FromParams` cannot tell if it was meant for the parent class, or if it's something used in the constructor of the child class.

tango/step_caches/remote_step_cache.py
AkshitaB: Potentially with beaker datasets

tango/integrations/beaker/common.py
```diff
@@ -3,97 +3,163 @@
...
+    def __init__(
+        self, beaker_workspace: Optional[str] = None, beaker: Optional[Beaker] = None, **kwargs
+    ):
+        user_agent = f"tango v{VERSION}"
+        self.beaker: Beaker
+        if beaker is not None:
+            self.beaker = beaker
+            if beaker_workspace is not None:
+                self.beaker.config.default_workspace = beaker_workspace
+                self.beaker.workspace.ensure(beaker_workspace)
```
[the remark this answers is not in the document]
AkshitaB: This is to ensure that the beaker object being used in the workspace and the step cache is the same. I've not changed this from the original implementation.

tango/common/remote_utils.py
AkshitaB: The clients are used in both step caches and workspaces. I still feel it's a useful abstraction to have, especially since we want to add stuff for search and filtering. We have control over beaker and its api, but not so much on gcloud/gcsfs. Having this layer helps with restricting changes in one place.

tango/integrations/beaker/executor.py
```diff
@@ -436,6 +440,17 @@ def __init__(
...
 
+        self.google_token = google_token or os.environ.get("GOOGLE_TOKEN")
+        # If the user passes file path instead of actual credentials dict, load the credentials
+        # so that they can be used in beaker.
+        if self.google_token is not None and self.google_token.endswith(".json"):
+            from tango.integrations.gs import GSWorkspace
+
+            if isinstance(workspace, GSWorkspace):
+                with open(self.google_token) as f:
+                    self.google_token = f.read()
```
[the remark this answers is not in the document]
AkshitaB: This is to maintain the functionality of being able to pass the credentials file path instead of the actual contents.

tango/integrations/gs/common.py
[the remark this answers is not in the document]
AkshitaB: This is the ridiculousness that I'm trying to debug right now:

The following code in Python Console:

```
import gcsfs
source = 'abc'  # this is a folder
target = 'allennlp-tango-bucket/folder'
fs = gcsfs.GCSFileSystem(consistency="md5")
fs.touch(target + "/.placeholder", truncate=True)
fs.put(source, target, recursive=True)
```

works perfectly - the files and folders in `abc` are present at `allennlp-tango-bucket/folder/file.txt`, `allennlp-tango-bucket/folder/subfolder` etc.

Put the EXACT same code in a python file and run that.

The folder `abc` is copied to the target, so that I end up with `allennlp-tango-bucket/abc`.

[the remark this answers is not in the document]
AkshitaB: I've opened an issue here: https://github.com/fsspec/gcsfs/issues/517. 
In the meantime, I've changed to a slightly cleaner workaround.

epwalsh:
I just ran into a weird issue while testing this out with the default executor. I started with an empty bucket and ran a single cacheable step. It doesn't look like my step was actually run:

image

I ran this several times with an empty bucket (same bucket, I just deleted everything before running again) and got the same result every time.

AkshitaB:
> I just ran into a weird issue while testing this out with the default executor. I started with an empty bucket and ran a single cacheable step. It doesn't look like my step was actually run:
> 
> image
> 
> I ran this several times with an empty bucket (same bucket, I just deleted everything before running again) and got the same result every time. If you check my GCS bucket "allennlp-petew", it doesn't look like the result of the step is in there.

@epwalsh Ah, I made an error in removing the concept of "commit". Will fix in a bit.

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,195 @@
...
+        if isinstance(step, (Step, StepInfo)):
+            cacheable = step.cache_results if isinstance(step, Step) else step.cacheable
+            if not cacheable:
+                return False
+
+            key = step.unique_id
+
+            # First check if we have a copy in memory.
+            if key in self.strong_cache:
+                print("found in strong!!!")
```
epwalsh, line 113: Left over from debugging?

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,195 @@
...
+                return True
+            if key in self.weak_cache:
+                print("found in weak!!!")
```
epwalsh, line 116: Same here

tango/step_caches/remote_step_cache.py
```diff
@@ -0,0 +1,195 @@
...
+                return True
+
+            # Then check if we have a copy on disk in our cache directory.
+            with self._acquire_step_lock_file(step, read_only_ok=True):
+                if self.step_dir(step).is_dir():
+                    print("found copy on disk!!!")
```
epwalsh, line 122: Same here

epwalsh:
Do you think the GCS workspace can efficiently implement the paginated workspace search methods proposed in https://github.com/allenai/tango/pull/489?

tango/step_caches/remote_step_cache.py
AkshitaB: Ugh, will remove these!

AkshitaB:
> Do you think the GCS workspace can efficiently implement the paginated workspace search methods proposed in #489?

@epwalsh Yes, Dirk and I discussed, and I'm looking at Firestore/Datastore for something like this.

epwalsh (approved):
LGTM! It already seems faster than the Beaker workspace. Do you think these abstractions will have to change if we integrate Firestore or whatever?

tango/integrations/gs/workspace.py
```diff
@@ -0,0 +1,225 @@
...
+          `~/.config/gcloud/application_default_credentials.json`.
+
+    """
+
+    Constants = Constants
+    NUM_CONCURRENT_WORKERS = 9  # TODO: increase and check
+
+    def __init__(
+        self,
+        workspace: str,
```
epwalsh, line 60: Wouldn't it make more sense to call this "bucket"?

epwalsh (approved):
This is cool!! And fast!

My only comments are about naming. I feel like we should call this integration "gcs" instead of "gs"? Or possibly "gc" for Google Cloud, since we're not just using GCS.. we're also using Datastore.

AkshitaB:
> My only comments are about naming. I feel like we should call this integration "gcs" instead of "gs"? Or possibly "gc" for Google Cloud, since we're not just using GCS.. we're also using Datastore.

@epwalsh My main reason for calling it "gs" rather than "gcs" was to keep things uniform with gsutil. "gs://workspace-bucket" is a valid location that people can copy-paste and use with cached-path or gsutil directly.

epwalsh:
Fair enough!

tango/integrations/gs/workspace.py
```diff
@@ -0,0 +1,225 @@
...
+            max_workers=self.NUM_CONCURRENT_WORKERS,
+            thread_name_prefix="GSWorkspace.registered_runs()-",
+        ) as executor:
+            run_futures = []
+            for run_entity in self._ds.query(kind="run").fetch():
+                run_futures.append(executor.submit(self._get_run_from_entity, run_entity))
+            for future in concurrent.futures.as_completed(run_futures):
+                run = future.result()
+                if run is not None:
+                    runs[run.name] = run
```
dirkgr, line 168: So run info is still in GCS, not in Datastore. If it's fast enough, even with many runs, that's fine. But it might be a good candidate to put into Datastore, too.

tango/integrations/gs/workspace.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def _update_step_info(self, step_info: StepInfo):
+
+        step_info_entity = self._ds.entity(
+            key=self._ds.key("stepinfo", step_info.unique_id),
+            exclude_from_indexes=("step_info_dict",),
+        )
+
+        # We can store each key separately, but we only index things that are useful for querying.
+        # TODO: do we want to index any other step_info keys?
```
dirkgr, line 209: Can't think of anything, but maybe it will become apparent when we merge the UI stuff?

dirkgr (changes requested):
It's entirely possible that I missed it and it's already there, but we need retries on all network calls, GCS and Datastore.

AkshitaB:
> It's entirely possible that I missed it and it's already there, but we need retries on all network calls, GCS and Datastore.

Cloud storage and Datastore both have retries already implemented. The default configuration is [here](https://cloud.google.com/storage/docs/retry-strategy#client-libraries) . Do you mean that we should have more control over the retries? We can customize them.

dirkgr:
> Cloud storage and Datastore both have retries already implemented. The default configuration is [here](https://cloud.google.com/storage/docs/retry-strategy#client-libraries) . Do you mean that we should have more control over the retries? We can customize them.

No, if we already have retries set up, this is fine. I like their default settings, too.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,372 @@
...
+        blobs = list(self.storage.bucket(self.bucket_name).list_blobs(prefix=path))
+        if len(blobs) > 0:
+            return self._convert_blobs_to_dataset(blobs)
+        else:
+            raise RemoteDatasetNotFound()
+
+    def create(self, dataset: str):
+        bucket = self.storage.bucket(self.bucket_name)
+        # gives refreshed information
+        if bucket.blob(os.path.join(dataset, self.placeholder_file)).exists():
```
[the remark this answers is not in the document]
AkshitaB: Good catch. Fixed.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,372 @@
...
+                    for filename in filenames:
+                        # TODO: CI fails with ThreadPoolExecutor parallelism. Debug later.
+                        source_file_path = os.path.join(dirpath, filename)
+                        target_file_path = os.path.join(
+                            folder_path, source_file_path.replace(source_path + "/", "")
+                        )
+                        _sync_blob(source_file_path, target_file_path)
+            else:
+                source_file_path = source_path
+                target_file_path = os.path.join(folder_path, os.path.basename(source_file_path))
```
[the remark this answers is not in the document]
AkshitaB: The source path in question is the local path, so I think this should be ok.

tango/integrations/gs/workspace.py
AkshitaB: Run info _is_ in datastore (`self._ds` is the datastore). The only run-related information stored in GCS is the log file.

tango/integrations/gs/common.py
```diff
@@ -0,0 +1,372 @@
...
+            folder_path = dataset.dataset_path
+        bucket = self.storage.bucket(self.bucket_name)
+        try:
+            bucket.delete_blob(os.path.join(folder_path, self.uncommitted_file))
+        except exceptions.NotFound:
+            if not bucket.blob(os.path.join(folder_path, self.placeholder_file)).exists():
+                raise RemoteDatasetNotFound()
+            # Otherwise, already committed. No change.
+
+    def download(self, dataset: GSDataset, target_dir: PathOrStr):
```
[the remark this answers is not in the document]
AkshitaB: This is already taken care of where the download is called: https://github.com/allenai/tango/blob/5426e32c8ca2f8f8a75153dff27236e2322f1877/tango/step_caches/remote_step_cache.py#L120

tango/common/remote_utils.py
```diff
@@ -0,0 +1,38 @@
...
+    STEP_ARTIFACT_PREFIX = "tango-step-"
+    STEP_INFO_FNAME = "step_info.json"
+    STEP_RESULT_DIR = "result"
+    STEP_GRAPH_ARTIFACT_PREFIX = "tango-step-graph-"
+    STEP_EXPERIMENT_PREFIX = "tango-step-"
+    STEP_GRAPH_FILENAME = "config.json"
+    GITHUB_TOKEN_SECRET_NAME: str = "TANGO_GITHUB_TOKEN"
+    RESULTS_DIR: str = "/tango/output"
+    INPUT_DIR: str = "/tango/input"
+    LOCK_ARTIFACT_SUFFIX: str = "-lock"
```
dirkgr, line 26: Why do some of these have types and others don't?

dirkgr (approved):
That is one big feature!

PR #419, opened 2022-09-23
Add `step_extra_dependencies` option to `Step` class

### LINKED ISSUE #418: Manual specification of step order

As far as I can tell (and I may be wrong about this), the only thing that dictates parent-child relations in the step graph is direct chaining of outputs to inputs via "ref". But it would be nice to have a way to _manually_ enforce that a step (S2) executes after another (S1), even if S1 is not an ancestor of S2 in the graph determined by "ref" relations. Is there a way to do this?

### Alternatives

A naïve solution to this would be to (e.g.) have the `run` method of S2 take the outputs of S1 as an argument and simply not use that argument. But this seems inelegant.

### Additional context

My personal use case motivating this is a set of LM prompting experiments in which one of the steps in the experiment simply prints a set of randomly sampled, dynamically constructed prompts. I would like to see what these prompts look like before model inference runs (as a separate step) in order to be confident the inputs are correct, but currently the prompts print only after the inference step. And the printing step and the model inference step are siblings in the step graph.

I am using the latest tango release (v0.14.0).

epwalsh:
Closes #418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added `step_requires` input field to `Step` class that can be used to force a dependency on another step even if the current step doesn't directly depend on the output of the other step. See [#418](https://github.com/allenai/tango/issues/418) for more context.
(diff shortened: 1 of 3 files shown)


dirkgr (changes requested):
This is fine, but can we call it something else, so there is no confusion about this vs. kwargs? I worry that someone will think they have to specify _all_ dependencies here. What if we call this "extra_dependencies" or something like that?

dirkgr (approved):

PR #430, opened 2022-09-29
Fix bug with Beaker executor

epwalsh:
The Beaker executor would hang at the end of a run if a step failed that was a dependency of another step.

This also makes the Beaker executor periodically print out updates so you know how many and which steps it's still waiting on.

### DIFF
--- a/tango/integrations/beaker/executor.py
+++ b/tango/integrations/beaker/executor.py
@@ -567,24 +567,24 @@ class BeakerExecutor(Executor):
             if now - last_progress_update >= 60 * 2:
                 last_progress_update = now
 
-                waiting_on = [
+                waiting_for = [
                     step_name
                     for step_name in submitted_steps
                     if step_name not in failed and step_name not in successful
                 ]
-                if len(waiting_on) > 5:
+                if len(waiting_for) > 5:
                     logger.info(
                         "Waiting on %d running steps...",
-                        len(waiting_on),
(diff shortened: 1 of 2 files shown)


tango/integrations/beaker/executor.py
```diff
@@ -515,6 +515,7 @@ def update_steps_to_run():
                                 # A dependency failed or can't be run, so this step can't be run.
                                 not_run[step_name] = ExecutionMetadata()
                                 steps_to_run.discard(step_name)
+                                steps_left_to_run.discard(step)
```
epwalsh, line 518: This line fixes the bug. The other additions below are just for the progress updates.

tango/integrations/beaker/executor.py
```diff
@@ -557,6 +558,51 @@ def future_done_callback(future: concurrent.futures.Future):
...
+                last_progress_update = now
+
+                waiting_on = [
+                    step_name
+                    for step_name in submitted_steps
+                    if step_name not in failed and step_name not in successful
+                ]
+                if len(waiting_on) > 5:
+                    logger.info(
+                        "Waiting on %d running steps...",
```
dirkgr, line 577: "Waiting for". If you're waiting on someone, you're bringing them water.

dirkgr (changes requested):
I like everything except the variable name 😅.

dirkgr (approved):

PR #437, opened 2022-10-04
Loosen range on some dependencies

epwalsh:


dirkgr (approved):
Big fan, especially before 1.0

PR #451, opened 2022-10-12
Use separate local cache dirs for each workspace

epwalsh:
Fixes issue where remote `Workspace` implementations would always use the same local cache directory, regardless of the workspace's configuration.

### DIFF
--- a/tango/common/util.py
+++ b/tango/common/util.py
@@ -187,7 +187,7 @@ def filename_is_safe(filename: str) -> bool:
 
 
 def make_safe_filename(name: str) -> str:
-    name = name.replace(" ", "-")
+    name = name.replace(" ", "-").replace("/", "--")
     return "".join(c for c in name if c in SAFE_FILENAME_CHARS)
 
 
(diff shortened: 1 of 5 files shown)


tango/common/util.py
```diff
@@ -186,6 +186,11 @@ def filename_is_safe(filename: str) -> bool:
     return all(c in SAFE_FILENAME_CHARS for c in filename)
 
 
+def make_safe_filename(name: str) -> str:
+    name = name.replace(" ", "-")
+    return "".join(c for c in name if c in SAFE_FILENAME_CHARS)
```
dirkgr, line 191: This can map multiple different `name`s to the same filename. Maybe use urlencode instead?

epwalsh: I don't think we'd want to encode this like a URL. You'd get URL escape codes in there which aren't file safe names. But we could append a hash of the original name to the resulting filename?

epwalsh: Also with the way we're using this it would be very unlikely to run into name clashes. So far we only use it to map W&B or Beaker workspace names, which themselves have character restrictions, to directories.

dirkgr: I'm OK with appending a hash. I realize collisions are unlikely, but time-consuming to debug when they happen. So I'd rather be safe.

epwalsh: Done: 49a965a

PR #462, opened 2022-11-07
Use new `constraints` field for cluster assignment

epwalsh:


codeviking:
I added @mewil, as he knows the constraint stuff better than I!

tango/integrations/beaker/executor.py
```diff
@@ -952,6 +915,7 @@ def _build_experiment_spec(
                 env_vars=self.env_vars,
                 priority=priority,
             )
+            .with_constraint("cluster", [clusters] if isinstance(clusters, str) else clusters)
```
mewil, line 918: Nice!

mewil (approved):

tango/integrations/beaker/executor.py
```diff
@@ -733,15 +697,19 @@ def _execute_sub_graph_for_step(
...
                     self.beaker.experiment.wait_for(
-                        experiment, strict=True, quiet=True, timeout=2.0
+                        experiment,
+                        strict=True,
+                        quiet=True,
+                        timeout=6.0,
+                        poll_interval=5.0,
                     )
-                    time.sleep(2.0)
+                    time.sleep(5.0)
```
epwalsh, line 712: Eases back the volume of requests FYI @codeviking

codeviking: So this will make a request every 5 seconds, instead of every 2, per job -- is that right?

This still feels like a lot. Most of our jobs are quite long. Plus, it often takes at least a minute or two for Beaker to start the job by shoveling around datasets, etc.

So I'd propose ratcheting this way back. What if we only asked for a job's status twice per minute?

aimichal: +1 to doing fewer requests for long-running jobs. Here's a potentially simple heuristic:

If time observing < 5 minutes: poll every 5 seconds
If time observing >= 5 minutes: poll ever 1 minute
If time observing >= 30 minutes: poll every 5 minutes

What do you think?

epwalsh: I think we can just keep it simple and always poll twice per minute

epwalsh: d08cbe9

dirkgr (approved):
Haven't run it yet, but code looks good.

PR #469, opened 2022-11-18
`BeakerExecutor` improvements, fix bug with `StepIndexer`

epwalsh:
Changes proposed:
- Adds field `run_local` to `BeakerExecutor`. You can use this to have the executor run certain steps locally. For instance, if you have a bunch of catwalk "tabulate-metrics" steps that are called "*-metrics" in your config, you could put `run_local: ["*-metrics"]` in your settings YAML file.
- Adds a field `gpu_type` to `StepResources`, and the `BeakerExecutor` will use this to find clusters with a matching GPU type.
- Fixes the unique ID for `StepIndexer`.

### DIFF
--- a/tango/step.py
+++ b/tango/step.py
@@ -67,6 +67,14 @@ class StepResources(FromParams):
     step to run.
     """
 
+    machine: Optional[str] = None
+    """
+    This is an executor-dependent option.
+
+    With the Beaker executor, for example, you can set this to "local" to force
+    the executor to run the step locally instead of on Beaker.
+    """
+
     cpu_count: Optional[float] = None
     """
     Minimum number of logical CPU cores. It may be fractional.
@@ -81,7 +89,11 @@ class StepResources(FromParams):
 
     gpu_type: Optional[str] = None
     """
-    The type of GPU that the step requires, e.g. 'NVIDIA A100-SXM-80GB'.
+    The type of GPU that the step requires.
+
+    The exact string you should use to define a GPU type depends on the executor.
+    With the Beaker executor, for example, you should use the same strings you
+    see in the Beaker UI, such as 'NVIDIA A100-SXM-80GB'.
     """
 
     memory: Optional[str] = None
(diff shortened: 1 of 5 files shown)


dirkgr:
The `run_local` thing is interesting. Why not make it a per-step setting though, so it's attached to the step instead of tango.yaml?

tango/step.py
```diff
@@ -79,6 +79,11 @@ class StepResources(FromParams):
     Minimum number of GPUs. It must be non-negative.
     """
 
+    gpu_type: Optional[str] = None
+    """
+    The type of GPU that the step requires, e.g. 'NVIDIA A100-SXM-80GB'.
```
dirkgr, line 84: Where do these strings come from? Can I see a list of them somewhere?

epwalsh:
> The run_local thing is interesting. Why not make it a per-step setting though, so it's attached to the step instead of tango.yaml?

That's probably a better way to do it. Maybe via `step_resources`?

dirkgr (approved):

tango/step.py
epwalsh: It depends on the executor. For the Beaker executor this should match what you see in the Beaker UI.

dirkgr:
> > The run_local thing is interesting. Why not make it a per-step setting though, so it's attached to the step instead of tango.yaml?
> 
> That's probably a better way to do it. Maybe via `step_resources`?

Yeah, some way of specifying less-than-nothing resources. "Don't even give me a machine for this. 0 CPUs."

epwalsh:
> > > The run_local thing is interesting. Why not make it a per-step setting though, so it's attached to the step instead of tango.yaml?
> > 
> > 
> > That's probably a better way to do it. Maybe via `step_resources`?
> 
> Yeah, some way of specifying less-than-nothing resources. "Don't even give me a machine for this. 0 CPUs."

I was thinking about adding another field, like "cluster" or "node" or "machine", which you'd set to "local".

dirkgr:
Yeah, that's probably better than the "0 CPU" thing :-)

PR #470, opened 2022-11-18
Remove tango server

epwalsh:


setup.py
```diff
@@ -103,7 +103,7 @@ def fix_url_dependencies(req: str) -> str:
         ],
     ),
     package_data={
-        "tango.server": ["*.svg", "*.css", "*.js", "*.html"],
+        #  "tango.server": ["*.svg", "*.css", "*.js", "*.html"],
```
dirkgr, line 106: Why is this just commented out?

dirkgr (changes requested):
It seems like you were super ready to take this thing out :-)

setup.py
epwalsh: We might need something like this again when we write the new server, so I figured I'd keep this here so I don't need to relearn how to do this in `setup.py`

dirkgr: I'm not in general a fan of that kind of thing, because you can say that about all sorts of things. Commented-out code is not a good place to keep some pointers-to-documentation for yourself.

The new server lives in its own repo, not here, so you won't see it when the time comes.

epwalsh: Oh, I was under the impression we'd be putting it here. In that case it's gone: 6aea54e

dirkgr (approved):

PR #471, opened 2022-11-22
reset sys.excepthook after importing other modules, add `--ext-var` CLI option

epwalsh:
Changes:
- [feature] adds a command line option to `tango run`: `--ext-var`. Use this to set JSONNET external variables when loading the experiment configuration file. For example, `--ext-var 'pretrained_model=gpt2'`.
- [bug fix]I was confused why our exception handling wasn't working as expected for me all of the sudden. After digging into it, I realized that `torchmetrics` (or some dependency of `torchmetrics`) is messing with `sys.excepthook`. But Tango should be only one messing with `sys.excepthook`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `gpu_type` field to `StepResources`. The `BeakerExecutor` can use this to determine which clusters to a submit a step to.
 - Added `machine` field to `StepResources`. You can set this to "local" when using the `BeakerExecutor` to force it to run the step locally.
+- Added `--ext-var` argument to `tango run` for setting JSONNET external variables
+  when loading the experiment config.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed issue where Executor `parallelism` option in a Tango settings file would be ignored.
 - Fixed a bug where the unique ID of a step that depends on a key-value of the result of another step could change if the name of the other step changes.
+- Fixed a bug where importing certain libraries (like torchmetrics) would mess with our exception handling because they set `sys.excepthook` for some reason. Now we always reset `sys.excepthook` after importing.
 
 ## [v1.0.2](https://github.com/allenai/tango/releases/tag/v1.0.2) - 2022-11-14
 
--- a/tango/__main__.py
+++ b/tango/__main__.py
@@ -286,7 +286,7 @@ def cleanup(*args, **kwargs):
     help="""Specify the name for this run.""",
 )
 @click.option(
-    "-e",
+    "-D",
     "--ext-var",
     type=str,
     help="""JSONNET external variables to use when loading the experiment config.
--- a/tango/__main__.py
+++ b/tango/__main__.py
@@ -80,7 +80,7 @@ import os
 import sys
 import warnings
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -20,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed issue where Executor `parallelism` option in a Tango settings file would be ignored.
 - Fixed a bug where the unique ID of a step that depends on a key-value of the result of another step could change if the name of the other step changes.
+- Fixed a bug where importing certain libraries (like torchmetrics) would mess with our exception handling because they set `sys.excepthook` for some reason. Now we always reset `sys.excepthook` after importing.
```
dirkgr, line 25: @ananyahjha93 Why does torchmetrics do this?

tango/__main__.py
```diff
@@ -285,6 +285,14 @@ def cleanup(*args, **kwargs):
...
     help="""Specify the name for this run.""",
 )
+@click.option(
+    "-e",
+    "--ext-var",
+    type=str,
+    help="""JSONNET external variables to use when loading the experiment config.
+    For example, --ext-var 'pretrained_model=gpt2'.""",
+    multiple=True,
+)
```
dirkgr, line 295: This is fine, but we could also import all environment variables into the Jsonnet context, right?

Can we make this a little more concise? `-D` for setting the variable, just like a C compiler? I don't care much about the long name, but `--ext-var` only makes sense when you know how Jsonnet thinks about the world, which is not something we should bother users with if we don't have to.

dirkgr (changes requested):

tango/__main__.py
epwalsh: > This is fine, but we could also import all environment variables into the Jsonnet context, right?

We already do this.

> Can we make this a little more concise?

Done: ec656cc

CHANGELOG.md
epwalsh: Digging into this a little more:

```python
import sys

print(sys.excepthook)  # <built-in function excepthook>

import torchmetrics  # 0.10.1

print(sys.excepthook)  # <function install.<locals>.excepthook at 0x12af553a0>
print(sys.excepthook.__module__)  # rich.traceback
```

epwalsh: But `rich` isn't a direct dependency of torchmetrics... And it's not a direct dependency of any of its direct dependencies either.. (numpy, torch, packing, typing_extensions)..

tango/__main__.py
epwalsh: But torchmetrics tries to import a bunch of dependencies that are not listed as dependencies, so it could be any of those.

CHANGELOG.md
epwalsh: But torchmetrics tries to import a bunch of dependencies that are not listed as dependencies, so it could be any of those.

dirkgr: I don't know either. But maybe they overwrite the except hook for a reason, and we're inviting trouble if we undo it?

epwalsh: I doubt that. The hook that's being set is this: https://rich.readthedocs.io/en/stable/traceback.html#traceback-handler

All it does is print a very verbose but pretty looking traceback. And in general I don't think we need to worry about not letting other libraries set the excepthook. I mean, what's the worst that could happen? It's only called when the program is crashing anyway...

dirkgr: Other libraries could be cleaning up resources, or do something else important.

I'm uneasy about this. Can we detect that it's `rich`, and only reset when that's the case?

epwalsh: Yeup: 9a6cd92

dirkgr (approved):

PR #474, opened 2022-11-23
Train is not optional

### LINKED ISSUE #472: Type of the argument `dataset` in `FlaxDataLoader` seems incorrect
The type of argument `dataset` in `FlaxDataLoader` is annotated as `DatasetDictBase`, but it does not appear to actually assume `DatasetDictBase`. First, it assumes the existence of the `num_rows` attribute or `"num_rows"` key for the `dataset`:

https://github.com/allenai/tango/blob/ccebb4c71dd5de097668cf652e4e2ddd684d2d25/tango/integrations/flax/data.py#L40-L42

Since `DatasetDictBase` seems to be the class that originally holds the split of the dataset, this operation should not be possible. Second, fancy indexing is applied to the `dataset`, which is also not an operation that can be performed on `DatasetDictBase`:

https://github.com/allenai/tango/blob/ccebb4c71dd5de097668cf652e4e2ddd684d2d25/tango/integrations/flax/data.py#L60

These actually make it look like a HuggingFace's `datasets.Dataset` is expected for the `dataset`. The unit test for `FlaxDataLoader` test actually only contains a test for `datasets.Dataset`.

To fix this, you would either need to simply annotate the `dataset` type with `datasets.Dataset`, or you would have to rewrite the `FlaxDataLoader` quite a bit.

### LINKED ISSUE #473: Type mismatch of `train_split` in `FlaxTrainStep`
The argument `train_split` of `FlaxTrainStep` is set as `Optional[str]` with the default value of `None`, but the documentation states that the default value is `"train"`:

https://github.com/allenai/tango/blob/ccebb4c71dd5de097668cf652e4e2ddd684d2d25/tango/integrations/flax/train.py#L85

https://github.com/allenai/tango/blob/ccebb4c71dd5de097668cf652e4e2ddd684d2d25/tango/integrations/flax/train.py#L120-L122

Since other `Step`s that also accept `train_split` set it as `train_split: str = "train"`, I thought this should be corrected as follows:

```diff
--- a/tango/integrations/flax/train.py
+++ b/tango/integrations/flax/train.py
@@ -82,7 +82,7 @@ class FlaxTrainStep(Step):
         seed: int = 42,
         keep_checkpoints: int = 5,
         lr_scheduler: Optional[Lazy[LRScheduler]] = None,
-        train_split: Optional[str] = None,
+        train_split: str = "train",
         validation_dataloader: Optional[Lazy[FlaxDataLoader]] = None,
         validation_split: Optional[str] = None,
         train_steps: Optional[int] = None,
```

dirkgr:
Fixes #473.
Fixed #472.


dirkgr:
I think that's all the changes I have.

dirkgr:
The flax example is still a mess. It seems to leak memory, and sometimes it doesn't use the GPU even though it's there and it says it's using it. But I can only fix so much of it, so I'll leave it here.

AkshitaB (approved):

PR #476, opened 2022-11-28
Add `@step()` decorator

epwalsh:
Adds a decorator for creating steps from functions:

```python
from tango import step

@step(version="001")
def add(a: int, b: int) -> int:
    return a + b
```


dirkgr:
😯

AkshitaB (approved):
Nice!

PR #489, opened 2022-12-09
Proposal: new paginated workspace search methods

epwalsh:


tango/workspace.py
```diff
@@ -70,6 +71,16 @@ def from_json_dict(cls, json_dict: Dict[str, Any]) -> "Run":
         return cls.from_params(params)
 
 
+class WorkspaceSort(StrEnum):
```
dirkgr, line 74: This is `RunSort`, no?

dirkgr: Or maybe these should just be called `Field`, since you could use the same names for filtering?

tango/integrations/beaker/workspace.py
```diff
@@ -331,25 +335,70 @@ def register_run(self, targets: Iterable[Step], name: Optional[str] = None) -> R
...
-
-        runs: Dict[str, Run] = {}
+    def search_registered_runs(
+        self,
+        *,
+        sort_by: WorkspaceSort = WorkspaceSort.START_DATE,
+        sort_descending: bool = True,
+        match: Optional[str] = None,
+        limit: Optional[int] = None,
+        cursor: Optional[int] = None,
```
dirkgr, line 345: Cursors are nice, but they promise a lot. They promise that the underlying list didn't change between calls. Maybe we should't make that kind of promise, because it'll be hard to keep.

tango/integrations/beaker/workspace.py
```diff
@@ -331,25 +335,70 @@ def register_run(self, targets: Iterable[Step], name: Optional[str] = None) -> R
...
-        import concurrent.futures
-
-        runs: Dict[str, Run] = {}
+    def search_registered_runs(
+        self,
+        *,
+        sort_by: WorkspaceSort = WorkspaceSort.START_DATE,
+        sort_descending: bool = True,
+        match: Optional[str] = None,
+        limit: Optional[int] = None,
```
dirkgr, line 344: Doesn't limit have to be a slice object, so you can request a range? Or is that how you use `cursor` below? In that case we should change the names. Cursors are something else.

dirkgr (changes requested):

tango/integrations/beaker/workspace.py
epwalsh: `cursor` and `limit` can be used together to get a slice. I named them this way to mirror Beaker's API, but we can call them something else.

tango/integrations/beaker/workspace.py
epwalsh: The intention of this API is not to make that promise, which I think is fine. As a consequence, on the UI the data for a given page might change as you flip back and forth through pages, but I think that's okay.

tango/workspace.py
epwalsh: I think `RunSort` makes sense. b5e7111

tango/integrations/beaker/workspace.py
dirkgr: I think those names are wrong in Beaker then too. That's not what a cursor is.

It's most Pythonic to do what the `slice` object does. `start`, `end`, `step`. And we don't really need `step` here.

tango/integrations/beaker/workspace.py
dirkgr: Yeah, while it would be super neat if we could support real cursors, it's not worth the effort.

tango/integrations/beaker/workspace.py
epwalsh: Okay, I'll make that change

epwalsh: Well, if we're not using `step`, why not just rename `cursor` to `start` and `limit` to `stop`. Like, what's the benefit of taking an actually `slice` object? I don't think it's Pythonic to have a method that takes a slice object, other than the magic method `__getitem__`.

epwalsh: 4ee9ba1

tango/workspaces/local_workspace.py
```diff
@@ -362,6 +362,38 @@ def registered_runs(self) -> Dict[str, Run]:
...
+        start: int = 0,
+        stop: Optional[int] = None,
+    ) -> List[StepInfo]:
+        with SqliteDict(self.step_info_file, flag="r") as d:
+            steps = [
+                step
+                for step in d.values()
+                if (match is None or match in step.unique_id)
+                and (state is None or step.state == state)
+            ]
```
dirkgr, line 381: I'm not a fan of how this has to read every step every time, but I guess there is no way around that. We'd have to change how these are stored.

dirkgr (approved):

tango/__init__.py
```diff
@@ -12,22 +12,27 @@
...
 )
-from tango.step import Step, StepResources, step
-from tango.step_cache import StepCache
-from tango.step_graph import StepGraph
-from tango.workspace import Workspace
+from .step import Step, StepResources, step
+from .step_cache import StepCache
+from .step_graph import StepGraph
+from .step_info import StepInfo, StepState
+from .workspace import RunSort, StepInfoSort, Workspace
```
dirkgr, line 38: Aren't relative imports a problem? I remember them breaking in various scenarios.

tango/integrations/gs/workspace.py
```diff
@@ -170,6 +173,170 @@ def registered_runs(self) -> Dict[str, Run]:
...
+        if sort_field is not None and not sort_locally:
+            order = [sort_field if not sort_descending else f"-{sort_field}"]
+
+        query = self._ds.query(kind="run", order=order)
+        if match is not None:
+            # HACK: Datastore has no direct string matching functionality,
+            # but we can assume that run names are alphanumeric. So this comparison
+            # is equivalent to checking if 'name' starts with 'match'.
+            query.add_filter("name", ">=", match)
+            query.add_filter("name", "<=", match + "~")
```
dirkgr, line 231: Do we have to escape `match` in some way? What if it already contains a `~`?

dirkgr (changes requested):
Requesting changes, but it might be nothing

tango/__init__.py
epwalsh: AFAIK the only place where relative imports break is within a `__main__` file, but there are other situations where absolute imports fail yet relative imports do not (https://bugs.python.org/issue25294). For the most part it just comes down to style / personal preference. I prefer using relative imports when importing things from submodules because it's more succinct, while using absolute imports when importing things from parent modules (or submodules of parent modules) because it's more clear. We've adopted that convention in most parts of this library (look at imports in any of our integrations, for example) but not everywhere (`step.py` doesn't use any relative imports).

tango/integrations/gs/workspace.py
epwalsh: I don't think we need to, assuming run names shouldn't have a `~` in them. But I suppose instead of using `~` to hack this, we could the `DEL` character (that's the last ASCII character). b883ad0

dirkgr: Oh, then wouldn't it be cleaner to do `..., "<", match[:-1] + chr(ord(match[-1]) + 1)`? Then it's not dependent on ASCII in any way.

tango/__init__.py
dirkgr: I would love to use relative imports if they didn't break `__main__`. But since they do, why is this safe? Are we saying, the only place they fail is within a `main()` function? So we can use them everywhere else?

I thought it was a problem with `FromParams`, because that relies on class identity, and relative imports make the same class appear under multiple different names.

tango/integrations/gs/workspace.py
epwalsh: Yes good point, although I think `match + chr(ord(match[-1]) + 1)` is fine instead of `match[:-1] + ...`.

epwalsh: 2e84840

tango/__init__.py
epwalsh: > I would love to use relative imports if they didn't break __main__. But since they do, why is this safe? Are we saying, the only place they fail is within a main() function? So we can use them everywhere else?

It doesn't have to do with the function, it's the module itself. To be more specific, what I meant is that the only place in `tango/` where relative imports won't work is `tango/__main__.py` when `tango` is run directly from the command-line (as opposed to running it as a module with `python -m tango ...`). This is because the module name at runtime of `tango/__main__.py` is `"__main__"`, so it doesn't act like it's inside of the `tango/` module, hence we have to use absolute imports there.

> I thought it was a problem with `FromParams`, because that relies on class identity, and relative imports make the same class appear under multiple different names.

I don't think that's true. It doesn't matter how you import a class, it's still the same class as far as Python is concerned.

epwalsh:
MyPy failure is unrelated and will be fixed by https://github.com/allenai/tango/pull/531.

tango/integrations/gs/workspace.py
dirkgr: > `match + chr(ord(match[-1]) + 1)` is fine instead of `match[:-1] + ...`

Really? So if `match` is `"cromulent"`, it'll be `< "cromulentu"`? I don't think that's correct. It should be `< "cromulenu"`.

dirkgr: Otherwise it will exclude the string `"cromulentz"`, which definitely as a prefix of `"cromulent"`.

epwalsh: Ah, you're right: 2857666

dirkgr (approved):

PR #502, opened 2023-01-09
Fix integration status messages in `tango info` command

epwalsh:
We shouldn't be trying to import every file in an integration's submodule to check if the integration is active, only the top-level submodule.


dirkgr (approved):
I was just thinking that we could do with some faster startup times in Tango ...

AkshitaB (approved):

PR #505, opened 2023-01-18
Rjsonnet

dirkgr:
This parses much faster than jsonnet.


epwalsh (approved):
Nice!

dirkgr:
I don't think this is going to work. Those test failures are genuine errors in the Rust code. I'm not going down that rabbit hole.

epwalsh:
Oddly "tests/common/params_test.py::TestParams::test_env_var_substitution" passes locally.

epwalsh:
I'd like to look into the issue more once I get a chance, so let's keep this open. I think it's still promising.

epwalsh:
I was able to reproduce the error outside of Tango. Opened an issue here: https://github.com/messense/rjsonnet-py/issues/6

dirkgr (commented):
Approve!

PR #509, opened 2023-01-19
Beaker-py upgrade in Beaker jobs

dirkgr:


tango/integrations/beaker/entrypoint.sh
```diff
@@ -30,6 +30,10 @@ fi
 # Configure git to use GitHub CLI as a credential helper so that we can clone private repos.
 gh auth setup-git
 
+# Every time Beaker changes their APIs, we need to upgrade beaker-py. This happens all the
+# time, so we make sure we have the latest.
+pip install --upgrade beaker-py
```
epwalsh, line 35: We should do this after activating the conda environment below

epwalsh (changes requested):

tango/integrations/beaker/entrypoint.sh
dirkgr: You are 100% right.

epwalsh (approved):

PR #516, opened 2023-01-28
Allow specifying multiple steps with `-s` from `tango run`

epwalsh:
For example, `tango run config.jsonnet -s foo -s bar`.

### DIFF
--- a/tango/__main__.py
+++ b/tango/__main__.py
@@ -775,7 +775,7 @@ def _run(
                 raise CliRunError
 
     if called_by_executor:
-        assert step_names
+        assert step_names is not None and len(step_names) == 1
 
         from tango.common.aliases import EnvVarNames
 
--- a/tango/__main__.py
+++ b/tango/__main__.py
@@ -263,8 +263,6 @@ def cleanup(*args, **kwargs):
     "-s",
     "--step-name",
     help="Execute a particular step (and its dependencies) in the experiment.",
-    type=str,
-    default=None,
     multiple=True,
 )
 @click.option(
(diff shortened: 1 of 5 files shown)


tango/__main__.py
```diff
@@ -750,15 +750,16 @@ def log_and_execute_run():
             cli_logger.info("[green]Starting new run [bold]%s[/][/]", run.name)
 
         executor_output: Optional[ExecutorOutput] = None
-        if step_name is not None:
+        if step_names:
```
dirkgr, line 753: That check isn't necessary. The for loop on line 755 won't run anyways if `step_names` is empty.

tango/__main__.py
```diff
@@ -774,12 +775,14 @@ def log_and_execute_run():
...
+        assert step_names
+
         from tango.common.aliases import EnvVarNames
 
         # We set this environment variable so that any steps that contain multiprocessing
         # and call `initialize_worker_logging` also log the messages with the `step_name` prefix.
-        os.environ[EnvVarNames.LOGGING_PREFIX.value] = f"step {step_name}"
-        initialize_prefix_logging(prefix=f"step {step_name}", main_process=False)
+        os.environ[EnvVarNames.LOGGING_PREFIX.value] = f"step {step_names[0]}"
+        initialize_prefix_logging(prefix=f"step {step_names[0]}", main_process=False)
```
dirkgr, line 785: Does this make sense? All the log messages have the prefix of the first step? That feels like it'll lead to confusion.

Why do we even have to say the step name in the prefix for the logs? Is it OK if we only do that when there is only one step, and otherwise we say "multiple steps", or no prefix at all?

dirkgr (changes requested):

tango/__main__.py
epwalsh: yea but the `else` branch here is important.

tango/__main__.py
epwalsh: This only happens when `called_by_executor` is `True`, in which case `len(step_names) == 1`. 2ef025b

dirkgr (approved):

PR #521, opened 2023-02-04
Shell step + Registrable functions

AkshitaB:
This PR adds a basic version of a shell step. 
It also makes adding registrable functions (so they can be referred in tango configs) simpler.

~TODO: separate out the registrable function stuff, as it's broader.~

### DIFF
--- a/tango/steps/shell_step.py
+++ b/tango/steps/shell_step.py
@@ -6,7 +6,7 @@ from tango.common import PathOrStr, Registrable
 from tango.step import Step
 
 
-class FunctionalRegistrable(Registrable):
+class RegistrableFunction(Registrable):
     WRAPPED_FUNC: ClassVar[Callable]
 
     def __call__(self, *args, **kwargs):
@@ -15,8 +15,8 @@ class FunctionalRegistrable(Registrable):
 
 def make_registrable(name: Optional[str] = None, *, exist_ok: bool = False):
     def function_wrapper(func):
-        @FunctionalRegistrable.register(name or func.__name__, exist_ok=exist_ok)
-        class WrapperFunc(FunctionalRegistrable):
+        @RegistrableFunction.register(name or func.__name__, exist_ok=exist_ok)
+        class WrapperFunc(RegistrableFunction):
             WRAPPED_FUNC = func
 
         return WrapperFunc
@@ -53,7 +53,7 @@ class ShellStep(Step):
         self,
         shell_command: str,
         output_path: Optional[PathOrStr] = None,
-        validate_output: FunctionalRegistrable = check_path_existence,
+        validate_output: RegistrableFunction = check_path_existence,
         **kwargs,
     ):
         output = self.run_command(shell_command, **kwargs)
--- a/tango/steps/shell_step.py
+++ b/tango/steps/shell_step.py
@@ -6,7 +6,7 @@ from tango.common import PathOrStr, Registrable
 from tango.step import Step
 
 
-class FunctionalRegistrable(Registrable):
+class RegistrableFunction(Registrable):
     WRAPPED_FUNC: ClassVar[Callable]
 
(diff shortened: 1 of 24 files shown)


tango/steps/shell_step.py
```diff
@@ -0,0 +1,73 @@
...
+
+def make_registrable(name: Optional[str] = None, *, exist_ok: bool = False):
+    def function_wrapper(func):
+        @RegistrableFunction.register(name or func.__name__, exist_ok=exist_ok)
+        class WrapperFunc(RegistrableFunction):
+            WRAPPED_FUNC = func
+
+        return WrapperFunc()
+
+    return function_wrapper
```
epwalsh, line 24: I think it would make more sense to put this in `tango/common/registrable.py`

tango/steps/shell_step.py
```diff
@@ -0,0 +1,73 @@
...
+
+    :param shell_command: The shell command to run.
+    :param output_path: The step makes no assumptions about the command being run. If your command produces some
+        output, ou can optionally specify the output path, for recording the output location, and optionally
+        validating it. See `validate_output` argument for this.
+    :param validate_output: If an expected `output_path` has been specified, you can choose to validate that the
+    step produced the correct output. By default, it will just check if the `output_path` exists, but you can
+    pass any other validating function. For example, if your command is a script generating a model output,
+    you can check if the model weights can be loaded.
+    :param kwargs: Other kwargs to be passed to the shell command.
```
epwalsh, line 49:
```suggestion
    :param kwargs: Other kwargs to be passed to `subprocess.run()`. Note that by default we
        set `shell = True`.
```

tango/steps/shell_step.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    def run(  # type: ignore[override]
+        self,
+        shell_command: str,
```
epwalsh, line 54: I think we should also accept a list of strings, since that's apparently [the preferred way](https://docs.python.org/3/library/subprocess.html#frequently-used-arguments):

> Providing a sequence of arguments is generally preferred, as it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names).

```suggestion
        shell_command: Union[str, List[str]],
```

tango/steps/shell_step.py
```diff
@@ -0,0 +1,73 @@
...
+        if output_path is not None:
+            validate_output(output_path)
+            self.logger.info(f"Output found at: {output_path}")
+
+        return str(output.decode("utf-8"))
+
+    def run_command(self, command: str, **kwargs):
+        self.logger.info("Command: " + command)
+        # shell=True to take advantage of environment variables, etc.
+        process = subprocess.run(command, capture_output=True, shell=True, **kwargs)
```
epwalsh, line 70: Maybe make this optional?
```suggestion
        # shell=True to take advantage of environment variables, etc.
        if "shell" not in kwargs:
            kwargs["shell"] = True
        process = subprocess.run(command, capture_output=True, **kwargs)
```

tango/steps/shell_step.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    This step runs a shell command, and returns the standard output as a string.
+
+    .. tip::
+
+        Registered as a :class:`~tango.step.Step` under the name "shell_step".
+
+    :param shell_command: The shell command to run.
+    :param output_path: The step makes no assumptions about the command being run. If your command produces some
+        output, ou can optionally specify the output path, for recording the output location, and optionally
```
epwalsh, line 43:
```suggestion
        output, you can optionally specify the output path for recording the output location, and optionally
```

epwalsh (changes requested):
Overall looks good! I have a few suggestions though

epwalsh (approved):
LGTM!

PR #527, opened 2023-02-09
Fix bug with variable length tuples

epwalsh:
Annotations like `Tuple[int, ...]` were not handled correctly in `FromParams`. You'd always get at most two elements, e.g. `[1, 2, 3, 4]` would be parsed into `[1, 2]`.


AkshitaB (approved):
Woah!

PR #536, opened 2023-03-06
Return more info from `Workspace.search_registered_runs()`

epwalsh:
@jonborchardt requested that we return as much info
as possible from this method. It makes the API a little
messy, but I think that's okay since these methods are
only designed for the UI.


dirkgr (approved):
It's not pretty, but in my experience web UIs always get into a state where you need APIs that return exactly what the UI needs...

PR #545, opened 2023-04-05
Only one step object

dirkgr:
#544 needs to be merged first.

We call `deepcopy()` on params objects, which contain Step objects, to set `raw_config`. That makes a copy of the step object every time it is used. Worse, this happens recursively, so we make a copy of a step object for every downstream step that depends on it directly or indirectly.

This solution replaces `Step` with refs when we're constructing the raw configs.

### DIFF
--- a/tango/step.py
+++ b/tango/step.py
@@ -364,7 +364,22 @@ class Step(Registrable, Generic[T]):
                     f"This happened when constructing an object of type {cls}."
                 )
 
-        raw_step_config = deepcopy(params.as_dict(quiet=True))
+        # Build up a raw step config
+        def replace_steps_with_refs(o: Any) -> Any:
+            if isinstance(o, (list, tuple, set)):
+                return o.__class__(replace_steps_with_refs(i) for i in o)
+            elif isinstance(o, (dict, Params)):
+                result = {key: replace_steps_with_refs(value) for key, value in o.items()}
+                if isinstance(o, dict):
+                    return result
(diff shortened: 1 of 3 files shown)


dirkgr:
It would not be pretty. We'd have to go through the arguments to the step and copy them manually, in the style of `_replace_step_dependencies()`. But it's possible.

epwalsh:
> It would not be pretty. We'd have to go through the arguments to the step and copy them manually, in the style of `_replace_step_dependencies()`. But it's possible.

Hmm, alright, I'll leave it up to you then. If you think this is the best way, that's fine with me.

tango/step.py
```diff
@@ -432,6 +432,14 @@ def from_params(  # type: ignore[override]
 
         return subclass(step_name=step_name, step_config=raw_step_config, **kwargs)
 
+    # We only want one step object per step with a given unique id.
+    def __copy__(self):
+        return self
+
+    # We only want one step object per step with a given unique id.
+    def __deepcopy__(self, memo=None):
```
AkshitaB, line 440: Can we put a log warning that this is being overridden? I don't like that the expected behavior is being changed quietly.

AkshitaB (commented):
Will this have unexpected behavior when step objects themselves contain some dictionaries, etc? Maybe won't matter as much when config files are used, but if someone is defining something programmatically?

AkshitaB (approved):

PR #573, opened 2023-06-05
LR schedulers

AkshitaB:


epwalsh (approved):
LGTM

I'm not sure what's up the failure in the GPU tests but that looks unrelated

PR #588, opened 2023-07-12
'Remove step' feature for local workspaces

pranjali97:


tango/workspace.py
```diff
@@ -419,6 +421,29 @@ def step_result(self, step_name: str) -> Any:
...
+                step_location = d[step_name].result_location
+                # remove step info from the sqlite dict
+                del d[step_name]
+                d.commit()
+                # remove cache directory
+                try:
+                    shutil.rmtree(step_location)
+                except OSError:
+                    raise OSError('Step Cache folder not found')
+                return('Step deleted')
```
AkshitaB, line 443: We don't really need to return a string. If no errors --> step successful.

tango/workspace.py
```diff
@@ -419,6 +421,29 @@ def step_result(self, step_name: str) -> Any:
                 return self.step_cache[run.steps[step_name]]
         raise KeyError(f"No step named '{step_name}' found in previous runs")
 
+
+    def remove_step(self, step_name: str) -> Any:
```
AkshitaB, line 425:
```suggestion
    def remove_step(self, step_name: str):
```

AkshitaB (commented):
I think the function in the workspace should be an abstract method: "NotImplementedError" , and this function is specifically for "local_workspace". 

The remote workspaces that inherit from local workspace can implement the remove method as follows:

```
def remove_step():
    super().remove_step()
    <remove specific step removal>
```

Also, we should add tests.

tango/workspace.py
pranjali97: sure! updated this!

tango/step_cache.py
```diff
@@ -48,6 +48,11 @@ def __setitem__(self, step: Step, value: Any) -> None:
         """Writes the results for the given step. Throws an exception if the step is already cached."""
         raise NotImplementedError()
 
+    @abstractmethod
+    def __delitem__(self, step_unique_id) -> None:
```
AkshitaB, line 52: We need type annotation here (and at other places that implement this function).

Also, let's have consistency in setitem and delitem. If setitem takes Step object as input, so should delitem.

tango/step_caches/local_step_cache.py
```diff
@@ -147,6 +160,14 @@ def __setitem__(self, step: Step, value: Any) -> None:
                 pass
             raise
 
+    def __delitem__(self, step_unique_id) -> None:
+        location = str(self.dir) + "/" + str(step_unique_id)
+        try:
+            shutil.rmtree(location)
+            assert not os.path.exists(location)
```
AkshitaB, line 167: Not necessary. Use the assert in the test instead.

tango/step_caches/memory_step_cache.py
```diff
@@ -35,6 +35,13 @@ def __setitem__(self, step: Step, value: Any) -> None:
                 UserWarning,
             )
 
+    def __delitem__(self, step_unique_id) -> None:
+        if step_unique_id in self.cache:
+            del self.cache[step_unique_id]
+            assert step_unique_id not in self.cache
```
AkshitaB, line 41: No need for assert.

tango/step_caches/memory_step_cache.py
```diff
@@ -35,6 +35,13 @@ def __setitem__(self, step: Step, value: Any) -> None:
...
+        else:
+            raise KeyError(f"{step_unique_id} not present in the memory cache. Can't be deleted")
```
AkshitaB, line 43:
```suggestion
            raise KeyError(f"{step_unique_id} not present in the memory cache. Cannot be deleted.")
```

tango/step_caches/remote_step_cache.py
```diff
@@ -155,3 +155,11 @@ def __setitem__(self, step: Step, value: Any) -> None:
 
         # Finally, add to in-memory caches.
         self._add_to_cache(step.unique_id, value)
+
+    def __delitem__(self, step_unique_id) -> None:
+        # check and delete local cache dir
+        if self.step_dir(step_unique_id).is_dir():
```
AkshitaB, line 161: We can use `super().__delitem__()` instead of replicating what we do in local cache. Keeps the local and remote logic separate. If we change local's implementation, we won't need to change remote's implementation.

tango/workspace.py
```diff
@@ -419,6 +419,14 @@ def step_result(self, step_name: str) -> Any:
                 return self.step_cache[run.steps[step_name]]
         raise KeyError(f"No step named '{step_name}' found in previous runs")
 
+    @abstractmethod
+    def step_cache_remove(self, step_unique_id: str):
```
AkshitaB, line 423: Let's just call this `remove_step`. Since we are not only removing the step from the cache, but also from step info, etc. Workspace object user does not need to know how/where steps are stored. All they care about is removing the step from the workspace.

tests/workspaces/local_workspace_cache_remove_test.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
AkshitaB, line 1: The test file name does not need to be specific to the one test we are currently running. We can just have a test file for `local_workspace_test.py`. This can be used for adding any other tests about the local workspace too (if needed, later). Similarly for other test files.

tests/workspaces/local_workspace_cache_remove_test.py
```diff
@@ -0,0 +1,24 @@
...
+from tango.step_info import StepState
+from tango.workspaces import LocalWorkspace
+
+
+class AdditionStep(Step):
+    def run(self, a: int, b: int) -> int:  # type: ignore
+        return a + b
+
+
+class Test_Cache_Remove_Workspace(TangoTestCase):
```
AkshitaB, line 14:
```suggestion
class TestLocalWorkspace(TangoTestCase):
```

General class name, also convention dictates no underscores in class names.

tests/workspaces/local_workspace_cache_remove_test.py
```diff
@@ -0,0 +1,24 @@
...
+    def test_step_cache_remove(self):
+        workspace = LocalWorkspace(self.TEST_DIR)
+        step = AdditionStep(a=1, b=2)
+        step_info = workspace.step_info(step)
+        assert step_info.state == StepState.INCOMPLETE
+        result = step.result(workspace)
+        step_info = workspace.step_info(step)
+        assert step_info.state == StepState.COMPLETED
+        step_unique_id = step.unique_id
+        workspace.step_cache_remove(step_unique_id)
```
AkshitaB, line 24: We should check that the step is no longer in the workspace. That's the main thing we are trying to test.
```suggestion
        workspace.step_cache_remove(step_unique_id)
        assert step not in workspace.step_cache
        # similarly for step_info
```

tests/workspaces/memory_workspace_cache_remove_test.py
```diff
@@ -0,0 +1,22 @@
...
+def test_step_cache_remove():
+    workspace = MemoryWorkspace()
+    step1 = AdditionStep(a=1, b=2)
+    step_info = workspace.step_info(step1)
+    assert step_info.state == StepState.INCOMPLETE
+    result1 = step1.result(workspace)
+    step_info = workspace.step_info(step1)
+    assert step_info.state == StepState.COMPLETED
+    step1_unique_id = step1.unique_id
+    workspace.step_cache_remove(step1_unique_id)
```
AkshitaB, line 22: Same as local workspace.

AkshitaB (changes requested):

tango/step_cache.py
AkshitaB, line 52:
```suggestion
    def __delitem__(self, step: Union[Step, StepInfo]) -> None:
```

AkshitaB: To match the implementations

tango/step_caches/local_step_cache.py
```diff
@@ -147,6 +160,14 @@ def __setitem__(self, step: Step, value: Any) -> None:
                 pass
             raise
 
+    def __delitem__(self, step: Union[Step, StepInfo]) -> None:
+        location = str(self.dir) + "/" + str(step.unique_id)
+        try:
+            shutil.rmtree(location)
+            self._remove_from_cache(step.unique_id)
+        except OSError:
+            raise OSError("Step Cache folder not found")
```
AkshitaB, line 169:
```suggestion
            raise OSError(f"Step cache folder for '{step.unique_id}' not found. Cannot be deleted.")
```

tango/step_caches/remote_step_cache.py
```diff
@@ -155,3 +155,6 @@ def __setitem__(self, step: Step, value: Any) -> None:
 
         # Finally, add to in-memory caches.
         self._add_to_cache(step.unique_id, value)
+
+    def __delitem__(self, step: Union[Step, StepInfo]) -> None:
+        super().__delitem__(step)
```
AkshitaB, line 160:
```suggestion

```

AkshitaB: Not required since it's inheriting the method anyway.

tests/integrations/beaker/workspace_test.py
```diff
@@ -22,3 +25,28 @@ def test_direct_usage(beaker_workspace: str):
...
+    assert workspace.beaker.dataset.get(dataset_name) is not None
+    assert step in cache
+
+    workspace.remove_step(step.unique_id)
+    cache = workspace.step_cache
+    dataset_name = workspace.Constants.step_artifact_name(step_info)
+
+    with pytest.raises(Exception) as excinfo:
+        workspace.beaker.dataset.get(dataset_name)
+    assert "DatasetNotFound" in str(excinfo)
```
AkshitaB, line 51:
```suggestion
    with pytest.raises(DatasetNotFound):
        workspace.beaker.dataset.get(dataset_name)
```

AkshitaB (commented):
Looks good overall. Added some minor suggestions. I can't reproduce the type check error locally, let's see the output of the checks after rerunning with the committed suggestions.

tests/integrations/beaker/workspace_test.py
AkshitaB: Ensure that it's imported

AkshitaB (approved):

PR #491, opened 2022-12-09
Add ability to display the executor output

BigRedT:
Fixes #487 

**Changes proposed in this pull request:**
- Lightly modified version of code [here](https://github.com/allenai/tango/blob/65f699d6b67a5afc05282cee1d190b8a30462328/tango/__main__.py#L797-L836)
Add the ability to display the executor output. Can be used as follows:

```python
output = executor.execute_step_graph(step_graph)
output.display()
```

**Tested locally and produces an output as follows:**

Screen Shot 2022-12-09 at 3 11 13 PM

## Before submitting

- [x] I've read and followed all steps in the [Making a pull request](https://github.com/allenai/tango/blob/main/CONTRIBUTING.md#making-a-pull-request)
    section of the `CONTRIBUTING` docs.
- [ ] I've updated or added any relevant docstrings following the syntax described in the
    [Writing docstrings](https://github.com/allenai/tango/blob/main/CONTRIBUTING.md#writing-docstrings) section of the `CONTRIBUTING` docs.
- [ ] If this PR fixes a bug, I've added a test that will fail without my fix.
- [ ] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [ ] All GitHub Actions jobs for my pull request have passed.

PR #600, opened 2023-09-09
CLI-style execution for Python-defined experiments

apoorvkh:
Hey!

I think the Tango CLI is really nice and would really like if we could support Python-defined experiments with CLI-style execution. So I modularized the CLI run functions and decoupled them from `__main__.py`.

Now we can also run scripts like:

```python
from tango import StepGraph
from tango.cli import tango_cli, execute_step_graph, prepare_workspace, prepare_executor

from examples.euler.complex_arithmetic import *
from tango.steps import PrintStep

i = [0.0, 1.0]
pi = [3.1415926535, 0.0]

i_times_pi = MultiplyStep(a=i, b=pi)
pow_e = ExponentiateStep(x=i_times_pi)
plus_one = AdditionStep(a=pow_e, b=[1,0])
print_step = PrintStep(input=plus_one)

step_graph = StepGraph({
    "i_times_pi": i_times_pi,
    "pow_e": pow_e,
    "plus_one": plus_one,
    "print": print_step
})

if __name__ == '__main__':
    with tango_cli():
        workspace = prepare_workspace()
        executor = prepare_executor(workspace)
        execute_step_graph(step_graph, workspace, executor)
```

This should be the same as `tango run examples/euler/euler.jsonnet`!

If we want to use the default `TangoGlobalSettings`, we can also shorten the block to

```python
    with tango_cli():
        execute_step_graph(step_graph)
```

This includes the changes from (@BigRedT) Tanmay's PR #491. CLI functionality should be completely the same. I ran the tests in `tests/main_test.py` successfully. Formatted with `isort` and `black` and checked with `ruff` and `mypy`. Docs needed.

What do you think? Happy to make adjustments as you see fit!


apoorvkh:
Seems like the lint and type check failed, but for reasons independent of this PR. Shall I resolve those anyway?

AkshitaB (approved):
@apoorvkh Thank you for the PR! The changes look good to me.

Some of the CI tests don't run due to permissions errors, but I ran some of them locally.

apoorvkh:
That's great -- thanks a lot @AkshitaB!

PR #601, opened 2023-09-18
CI errors be gone

AkshitaB:


AkshitaB:
Docker build is failing because we need torch 2.0.1 for the `LRScheduler` bug fix on their end, and the `Dockerfile` starts with a base image of torch 2.0.0

epwalsh (approved):
LGTM

PR #604, opened 2023-10-04
GS workspaces can be bucket subfolders

### LINKED ISSUE #603: GS workspace should be able to use bucket subfolders
Currently, it appears that the GS workspace needs to be mapped to a bucket. We would like to be able to use `gs://bucket-name/my-workspace` etc. also.

AkshitaB:
Fixes #603


epwalsh (approved):
LGTM!

PR #607, opened 2023-10-20
GS workspace bug fixes

AkshitaB:
Multiple concurrent datastore get calls caused the tango run to hang. Instead, we now use datastore's method to get step info of multiple steps.


epwalsh (approved):
LGTM

